PR #6, opened 2020-05-06
Split into two crates: psa-crypto-sys and psa-crypto.

egrimley-arm:
psa-crypto-sys exposes a minimal low-level C interface to Rust.
psa-crypto will be a higher-level, more Rust-friendly interface.


psa-crypto-sys/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+    non_snake_case,
+    non_camel_case_types,
+    non_upper_case_globals,
+    dead_code,
+    trivial_casts
+)]
+#[allow(clippy::all)]
+mod psa_crypto_binding {
+    include!(concat!(env!("OUT_DIR"), "/shim_bindings.rs"));
+}
```
ionut-arm, line 54: Any reason why not just doing the `include!` here?

egrimley-arm: In `shim_bindings.rs` there are a load of symbols derived from transitively included C headers. There's a slight risk one of them might collide with a symbol we're using in Rust. I would guess that's a justification for doing it this way, though I just copied how it was done previously.

hug-dev: Oh yes, that is a good point, we can re-export from this crate only the ones defined in the PSA Crypto API itself.

PR #7, opened 2020-05-06
Use "development" branch from git repo instead of old tar file.

egrimley-arm:
This would fix parallaxsecond/parsec#153.

Instead of
wget https://github.com/ARMmbed/mbed-crypto/archive/mbedcrypto-2.0.0.tar.gz
we fetch the Mbed Crypto source with:
git clone https://github.com/ARMmbed/mbedtls && git checkout $MBED_COMMIT

The MBED_COMMIT is specified in Cargo.toml and will be updated from
time to time at least until there is a stable branch we can use.

Two functions in the API have changed their names. These are renamed
in psa-crypto-sys, not in psa-crypto.

### DIFF
--- a/psa-crypto-sys/Cargo.toml
+++ b/psa-crypto-sys/Cargo.toml
@@ -20,4 +20,4 @@ serde = { version = "1.0", features = ["derive"] }
 toml = "0.4.2"
 
 [package.metadata.config]
-mbed-crypto-version = "mbedcrypto-2.0.0"
+mbed-crypto-commit = "5435451a1a3d0af3eef71e3883f3087df52e48d5"
--- a/psa-crypto-sys/build.rs
+++ b/psa-crypto-sys/build.rs
@@ -37,7 +37,7 @@ use std::io::{Error, ErrorKind, Result};
 use std::path::{Path, PathBuf};
 
 const CONFIG_TABLE_NAME: &str = "config";
-const MBED_CRYPTO_VERSION_KEY: &str = "mbed-crypto-version";
+const MBED_CRYPTO_COMMIT_KEY: &str = "mbed-crypto-commit";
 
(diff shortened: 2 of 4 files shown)


hug-dev:
> I don't know how to make PR on top of an existing PR

I don't think you can 😋 

About the PR, thanks for the changes, we should definitely use the development branch of Mbed TLS.
I have been reading those two articles ([1](https://s3.amazonaws.com/temp.michaelfbryan.com/wrap-libmagic/index.html), [2](https://kornel.ski/rust-sys-crate)) and I think we should follow what they prescribe:
1. Have a folder `vendor` where `mbedtls` is put a a git submodule
2. If the environment variables `MBEDTLS_LIB_DIR` and `MBEDTLS_INCLUDE_DIR` are not set, compile the library from the sources in the submodule folder (into `OUT_DIR`). In the future we would use `pkg-config` to find if the library is installed on the system as well ([when it is supported](https://github.com/ARMmbed/mbedtls/pull/1753))
3. All of our compilation logic should be in the `build.rs` file and not using scripts I think. We ought to use available crates for compilation like `make` or `cc`.
4. Create `static` and `dynamic` features to choose between the two but when compiling locally the library, link statically.

ionut-arm:
> 3. All of our compilation logic should be in the build.rs file and not using scripts I think. We ought to use available crates for compilation like `make` or `cc`.

I agree with this in principle, but sometimes it's just not feasible. The [`make`](https://crates.io/crates/make) crate, for example, is at version 0.0.0 and has no documentation. Didn't dig too much but couldn't see anything else related. [`cc`](https://crates.io/crates/cc) is in a much better state.

hug-dev:
The `psa-crypto-sys` crate now builds Mbed TLS from source, currently using commit `6abc20e`. @egrimley-arm is that fine for you if we close this PR?

egrimley-arm:
Yes, let's close it.

PR #10, opened 2020-05-19
Split between Status and Error

hug-dev:
Also implements conversions between FFI types and Rust native types.

### DIFF
--- a/psa-crypto/src/types/status.rs
+++ b/psa-crypto/src/types/status.rs
@@ -8,13 +8,20 @@
 use log::error;
 
 /// Result type returned by any PSA operation
-pub type Result<T> = core::result::Result<T, Status>;
+pub type Result<T> = core::result::Result<T, Error>;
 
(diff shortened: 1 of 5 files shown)


psa-crypto/src/types/status.rs
```diff
@@ -19,6 +19,16 @@ pub enum Status {
...
 
+impl Status {
+    /// Convert the Status into a Result returning the empty tuple
+    pub fn to_result(self) -> Result<()> {
+        match self {
+            Status::Success => Ok(()),
+            Status::Error(error) => Err(error),
+        }
+    }
+}
```
ionut-arm, line 30: Why not implementing `From<Status> for Result`?

hug-dev: I will add it as well! I am keeping both as in some contexts when the compiler can not infer the type it might be more convenient to use `...to_result()` rather than `Result<()>::from(...)`

PR #12, opened 2020-05-21
Change dependency to MbedTLS and update build

ionut-arm:
This commit move the dependency of the crate to MbedTLS, where Mbed
Crypto was moved. It also moves all build steps within the `build.rs`
file.

Co-authored-by: Edmund Grimley Evans <edmund.grimley-evans@arm.com>
Co-authored-by: Ionut Mihalcea <ionut.mihalcea@arm.com>


psa-crypto-sys/build.rs
```diff
@@ -195,84 +98,49 @@ fn generate_mbed_bindings(mbed_config: &MbedConfig, mbed_version: &str) -> Resul
...
-        // Request rustc to link the Mbed Crypto static library
-        println!(
-            "cargo:rustc-link-search=native={}/mbed-crypto-{}/library/",
-            mbed_config
-                .mbed_path
-                .unwrap_or_else(|| env::var("OUT_DIR").unwrap()),
-            mbed_version,
-        );
-        println!("cargo:rustc-link-lib=static=mbedcrypto");
+fn main() -> Result<()> {
```
hug-dev, line 127: The logic is much simpler and nicer!
I have been reading [the article](https://kornel.ski/rust-sys-crate) again and I think we should follow this decision path:

1. Is the library available? (We should ideally use `pkg-config` for that but currently use the env variable as you did). "Yes" go to 2, "No" go to 3.
2. Link with the library _dynamically_ by default or statically if the `static` feature is set
3. Compile the library from the submodule and link with it **always statically**.

I think that libraries that we compile ourselves should always be linked statically but we leave the option for libraries that are available on the system.

psa-crypto-sys/build.rs
```diff
@@ -29,155 +29,58 @@
...
-    };
-
-    println!("cargo:rerun-if-changed={}", SETUP_MBED_SCRIPT_PATH);
-    println!("cargo:rerun-if-changed={}", "src/c/Makefile");
+fn setup_mbed_crypto() -> Result<()> {
+    let mbedtls_dir = String::from("./mbedtls");
     println!("cargo:rerun-if-changed={}", "src/c/shim.c");
     println!("cargo:rerun-if-changed={}", "src/c/shim.h");
+    // The SHARED env variable tells the build to output dynamic libraries too
+    env::set_var("SHARED", "1");
```
hug-dev, line 41: Following the "decision path" comment below, this would not be needed if we always link with the libs we compile ourselves statically.

PR #16, opened 2020-05-26
Add documentation, examples and tests

hug-dev:


psa-crypto/src/operations/asym_signature.rs
```diff
@@ -2,13 +2,57 @@
...
+/// #     policy: Policy {
+/// #         usage_flags: UsageFlags {
+/// #             sign_hash: true,
+/// #             sign_message: true,
+/// #             verify_hash: true,
+/// #             verify_message: true,
+/// #             ..Default::default()
+/// #         },
+/// #         permitted_algorithms: AsymmetricSignature::RsaPkcs1v15Sign {
+/// #             hash_alg: Hash::Sha256.into(),
```
egrimley-arm, line 37: Why do this line and the following line contain `#`?

psa-crypto/src/operations/asym_signature.rs
```diff
@@ -38,7 +82,49 @@ pub fn sign_hash(
     Ok(signature_length)
 }
 
-/// Verify a hash
+/// Verify the signature of a hash or short message using a public key
+///
+/// # Example
+///
+/// ```
+/// # use psa_crypto::operations::key_management::generate;
```
egrimley-arm, line 90: Ditto.

psa-crypto/src/types/key.rs
```diff
@@ -35,6 +35,39 @@ impl Attributes {
     }
 
     /// Check export in a faillible way
```
egrimley-arm, line 37: Not part of this patch, but while you're there: faillible -> fallible

psa-crypto/src/operations/asym_signature.rs
hug-dev: It is made to hide portions of the example so it only shows relevent parts. See [here](https://doc.rust-lang.org/rustdoc/documentation-tests.html#hiding-portions-of-the-example) for details.

psa-crypto/src/types/algorithm.rs
```diff
@@ -42,6 +42,17 @@ pub enum Algorithm {
...
 impl Algorithm {
     /// Check if the algorithm is a HMAC algorithm, truncated or not
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use psa_crypto::types::algorithm::{Algorithm, Mac, FullLengthMac, Hash};
+    /// let hmac = Algorithm::Mac(Mac::Truncated {
+    ///     mac_alg: FullLengthMac::Hmac { hash_alg: Hash::Sha256 },
+    ///     mac_length: 34,
```
ionut-arm, line 52: I'm assuming the `mac_length` is in bytes, not bits, in which case this is invalid because `Sha256` has only 32 bytes (not that I'm saying we shouldn't allow that construction, just saying that maybe something below 32 would be better for an example)

psa-crypto/src/types/algorithm.rs
```diff
@@ -685,3 +720,42 @@ impl From<AsymmetricSignature> for psa_crypto_sys::psa_algorithm_t {
...
+                .unwrap()
+        );
+        assert_eq!(
+            psa_crypto_sys::PSA_ALG_ECDSA(psa_crypto_sys::PSA_ALG_SHA3_512),
+            Algorithm::AsymmetricSignature(AsymmetricSignature::Ecdsa {
+                hash_alg: SignHash::Specific(Hash::Sha3_512),
+            })
+            .try_into()
+            .unwrap()
+        );
```
ionut-arm, line 759: Could you add some checks that fail (i.e. with `unwrap_err()`)?

psa-crypto/src/types/key.rs
```diff
@@ -390,6 +469,23 @@ pub struct UsageFlags {
...
+            encrypt: false,
+            decrypt: false,
+            sign_message: false,
+            verify_message: false,
+            sign_hash: false,
+            verify_hash: false,
+            derive: false,
+        }
+    }
+}
```
ionut-arm, line 487: Is this not what deriving `Default` would obtain?

psa-crypto/src/types/key.rs
```diff
@@ -1017,4 +1114,63 @@ mod tests {
...
+                    },
+                    permitted_algorithms: AsymmetricSignature::Ecdsa {
+                        hash_alg: Hash::Sha256.into(),
+                    }
+                    .into(),
+                },
+            },
+            attrs.try_into().unwrap()
+        );
+    }
```
ionut-arm, line 1175: 👌 nice

psa-crypto/src/operations/key_management.rs
```diff
@@ -8,8 +8,44 @@ use crate::types::key::{Attributes, Id};
 use crate::types::status::{Result, Status};
 use core::convert::TryFrom;
 
-/// Generate a key
-pub fn generate_key(attributes: Attributes, id: Option<u32>) -> Result<Id> {
```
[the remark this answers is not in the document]
hug-dev: > How come you dropped the _key part of the name, because it's in key_management? 

Yes, that's the reason!

> Then you probably need to modify the examples to call it as key_management::generate(...), because it's confusing as to what generate is and where it is coming from.

Sure, will do.

psa-crypto/src/types/algorithm.rs
hug-dev: Makes sense, will modify!

psa-crypto/src/types/algorithm.rs
hug-dev: Yes, will add!

psa-crypto/src/types/key.rs
hug-dev: Oh yes, you are right, I forgot that you can derive `Default`

psa-crypto/src/operations/key_management.rs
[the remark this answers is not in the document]
ionut-arm: You've only modified the examples in this file, the other still import the function directly 🤡 but nevermind, we can discuss under #17

PR #19, opened 2020-06-01
Add no-std feature; impl Display on Error

ionut-arm:
This commit adds a no-std feature that controls whether the crate
exposes any functionality that makes use of the standard library. The
feature is enabled by default.

In "std"-mode, the Display and Error traits are implemented on the
`Error` enum.

### DIFF
--- a/psa-crypto/Cargo.toml
+++ b/psa-crypto/Cargo.toml
@@ -19,5 +19,6 @@ log = "0.4.8"
 serde = { version = "1.0.110", features = ["derive"] }
 
 [features]
-default = ["with-mbed-crypto"]
+default = ["with-mbed-crypto", "no-std"]
 with-mbed-crypto = ["psa-crypto-sys"]
+no-std = []
--- a/psa-crypto/src/types/status.rs
+++ b/psa-crypto/src/types/status.rs
@@ -8,6 +8,9 @@
 #[cfg(feature = "with-mbed-crypto")]
 use log::error;
 
+#[cfg(not(feature = "no-std"))]
+use std::fmt;
+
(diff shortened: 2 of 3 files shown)


psa-crypto/src/types/status.rs
```diff
@@ -89,6 +94,73 @@ pub enum Error {
...
+            Error::InvalidHandle => write!(
+                f,
+                "The key handle is not valid"
+            ),
+        }
+    }
+}
+
+#[cfg(not(feature = "no-std"))]
+impl ErrorTrait for Error {}
```
hug-dev, line 162: You could directly do:
```rust
impl std::error::Error for Error {}
```
to avoid the renaming and the `use` above. Leaving it at your choice, I do not mind!

psa-crypto/Cargo.toml
```diff
@@ -19,5 +19,6 @@ log = "0.4.8"
 serde = { version = "1.0.110", features = ["derive"] }
 
 [features]
-default = ["with-mbed-crypto"]
+default = ["with-mbed-crypto", "no-std"]
 with-mbed-crypto = ["psa-crypto-sys"]
+no-std = []
```
hug-dev, line 24: How would one compile this crate with only the `no-std` feature activated? Would it be like:
```
cargo build --no-default-features --features no-std
```

ionut-arm: Yes, precisely!

psa-crypto/src/types/status.rs
ionut-arm: Done!

PR #20, opened 2020-06-01
Add various improvements

hug-dev:
This commit:
* fixes a bug in build.rs where the shim library was not compiled when
dynamically linking
* add the psa_get_key_id function to the shim
* allow to access psa-crypto-sys from psa-crypto if compiling with Mbed
Crypto
* add a conversion from Error to psa_status_t to avoid going to Status

### DIFF
--- a/psa-crypto-sys/Cargo.toml
+++ b/psa-crypto-sys/Cargo.toml
@@ -15,7 +15,7 @@ repository = "https://github.com/parallaxsecond/rust-psa-crypto"
 links = "mbedcrypto"
 
 [build-dependencies]
-bindgen = "0.50.0"
+bindgen = "0.54.0"
 cc = "1.0.54"
 cmake = "0.1.44"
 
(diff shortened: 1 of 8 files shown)


psa-crypto-sys/Cargo.toml
```diff
@@ -15,7 +15,7 @@ repository = "https://github.com/parallaxsecond/rust-psa-crypto"
 links = "mbedcrypto"
 
 [build-dependencies]
-bindgen = "0.50.0"
+bindgen = "0.54.0"
```
ionut-arm, line 18: 💯

psa-crypto/src/types/status.rs
```diff
@@ -162,6 +162,39 @@ impl From<Status> for psa_crypto_sys::psa_status_t {
     }
 }
 
+#[cfg(feature = "with-mbed-crypto")]
+impl From<Error> for psa_crypto_sys::psa_status_t {
```
ionut-arm, line 166: If you have this then you don't need to have the whole match again above, in the `From<Status> for psa_status_t`, you can just do `Status::Error(error) => error.into()`

hug-dev: Good point! Will push this modification.

ionut-arm: Thanks! This was actually one of the things that I needed for the `ResponseStatus` update :)

PR #23, opened 2020-06-03
Add conversion from psa_status_t to Error

ionut-arm:
This commit adds a conversion from psa_status_t to Error, making the
conversion to Status use this one underneath.

### DIFF
--- a/psa-crypto/src/types/status.rs
+++ b/psa-crypto/src/types/status.rs
@@ -164,32 +164,40 @@ impl From<Error> for Status {
     }
 }
 
+impl From<psa_crypto_sys::psa_status_t> for Error {
+    fn from(status: psa_crypto_sys::psa_status_t) -> Self {
+        match status {
+            psa_crypto_sys::PSA_ERROR_GENERIC_ERROR => Error::GenericError,
+            psa_crypto_sys::PSA_ERROR_NOT_SUPPORTED => Error::NotSupported,
+            psa_crypto_sys::PSA_ERROR_NOT_PERMITTED => Error::NotPermitted,
+            psa_crypto_sys::PSA_ERROR_BUFFER_TOO_SMALL => Error::BufferTooSmall,
+            psa_crypto_sys::PSA_ERROR_ALREADY_EXISTS => Error::AlreadyExists,
+            psa_crypto_sys::PSA_ERROR_DOES_NOT_EXIST => Error::DoesNotExist,
+            psa_crypto_sys::PSA_ERROR_BAD_STATE => Error::BadState,
+            psa_crypto_sys::PSA_ERROR_INVALID_ARGUMENT => Error::InvalidArgument,
+            psa_crypto_sys::PSA_ERROR_INSUFFICIENT_MEMORY => Error::InsufficientMemory,
+            psa_crypto_sys::PSA_ERROR_INSUFFICIENT_STORAGE => Error::InsufficientStorage,
+            psa_crypto_sys::PSA_ERROR_COMMUNICATION_FAILURE => Error::CommunicationFailure,
+            psa_crypto_sys::PSA_ERROR_STORAGE_FAILURE => Error::StorageFailure,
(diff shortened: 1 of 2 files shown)


psa-crypto/src/types/status.rs
```diff
@@ -164,36 +164,44 @@ impl From<Error> for Status {
     }
 }
 
-impl From<psa_crypto_sys::psa_status_t> for Status {
+impl From<psa_crypto_sys::psa_status_t> for Error {
```
hug-dev, line 167: Is it necessary to have this conversion if we have already the one for `Status`? I am worried about it because have the one before was good logically as `Status` is the Rust equivalent of `psa_status_t`. With this one we could have a conversion from `PSA_SUCCESS` to `Error:GenericError`.
To have an `Error`, could you first convert to a `Status` and then match?

ionut-arm: I definitely agree with that and it is indeed wrong to map `PSA_SUCCESS` to `GenericError`, but I think the way the status type is defined is very C'ish and somewhat unsound in a more rich language like Rust. 

My view is that `Status` isn't that much representative of the state of the library/code/whatever like its name would imply, it's only purpose is to define types of failures, if they exist. This is like an anti-Null - you either have an error when something went wrong, or you get a "Null" value, 0, "success". So `Status` isn't really a valid characterisation because it doesn't tell you anything about a valid state, it's actually `Option<Error>`... I can drop this PR since, as you've said, in a way the conversion to `Status` makes more sense, but I'm just saying that `Status` is logically flawed from a Rust perspective.

ionut-arm: (even the C code would make more sense, if it could use `Option<psa_error_t>`)

ionut-arm: Actually, I think that's a good point - why do we even have both `Status` and `Result<T, Error>` since they really mean the same thing, while `Result` is actually more idiomatic, has more functionality for dealing with the two cases (`map`, `unwrap` etc) and can actually hold any type for return value.

ionut-arm: And this applies equally well to `ResponseStatus` - it's not idiomatic, that's why we have to migrate to `Result<T, Error>` instead of `Result<T, ResponseStatus>`

hug-dev: I think it is fine to have the `Status` and `ResponseStatus` types defined as they are nice to convert with their integer equivalents but totally agree that it is not idiomatic to have them in the `Err` variant of a `Result`.
I would say it is fine for these types to exist because they are useful in some cases and have a logical value but we should not use them in `Result`.
I think it is idiomatic to have `Status` but not `Result<T, Status>` and same for `ResponseStatus`.

ionut-arm: > Status and ResponseStatus types defined as they are nice to convert with their integer equivalents

Then they should not be public, if their only use is for conversion, otherwise what's the purpose of them? They're not included in ANY interface, not needed anywhere in the client or for anyone writing any code outside of the "interface" (`parsec-interface` or actually inside the `psa-crypto` crate) so making them public is both not needed and possibly confusing. In which case - do we really need them? Do we need a type just for a conversion to look nice?

hug-dev: I think we created `Status` just to follow the idiom of one Rust native structure per FFI type. I think it is nice to have such mapping and would say that people already knowing PSA Crypto API could easily map each of its type to one that we have in this crate, in the docs.

Agree that when using this crate from Rust to C this type does not need to be used. However when going the other way around, when developping a Secure Element Driver for example, it is nice to first convert to a `Status` and then let this crate do the conversion from `Status` to `psa_status_t`. I think it is nicer (and safer) for SE drivers to implement the conversion from `T` to `Status` rather than from `T` to `psa_status_t`.

ionut-arm: Fair enough, I think the SE driver use-case makes sense! But still unsure about `ResponseStatus` since that's really not useful outside of the IPC boundary

PR #27, opened 2020-06-15
Modify Mbed TLS version to most recent realease

hug-dev:
Last release of Mbed TLS included Nbed Crypto so we use that instead of
a commit number.
Also adds tests for dynamic linking on the CI.


hug-dev:
Just realised that the fix in Mbed TLS ARMmbed/mbedtls#3367 is not part of the realease 🤡 Which means that this repo, with `2.22.0`, can not be used and published in a crate.

Issue #26, opened 2020-06-10
Always close key handle if key is persistent


sbailey-arm:
Currently handles for persistent keys are kept open after generating and importing, which causes mbedtls to run out of key slots.

If a key is persistent, its handle should be closed after each generate and import operation.

Only set the `handle` field if the key is volatile.

hug-dev:
Done in linked PR

PR #28, opened 2020-06-15
Part of moving Parsec to use psa-crypto

sbailey-arm:
Part of  [Parsec Issue #177](https://github.com/parallaxsecond/parsec/issues/177)
Includes #26 

Contains changes required to allow [Prasec](https://github.com/parallaxsecond/parsec) to use `psa-crypto` instead of binding directly to `mbedtls`.

### DIFF
--- a/psa-crypto-sys/src/c/shim.h
+++ b/psa-crypto-sys/src/c/shim.h
@@ -72,33 +72,21 @@ const psa_algorithm_t shim_PSA_ALG_TLS12_PSK_TO_MS_BASE =
 const psa_algorithm_t shim_PSA_ALG_KEY_DERIVATION_MASK =
     PSA_ALG_KEY_DERIVATION_MASK;
 
-psa_key_id_t shim_get_key_id(const psa_key_attributes_t *attributes);
-
-size_t shim_get_key_bits(const psa_key_attributes_t *attributes);
-
-psa_key_type_t shim_get_key_type(const psa_key_attributes_t *attributes);
-psa_key_lifetime_t
-shim_get_key_lifetime(const psa_key_attributes_t *attributes);
 psa_algorithm_t shim_get_key_algorithm(const psa_key_attributes_t *attributes);
-psa_key_usage_t
-shim_get_key_usage_flags(const psa_key_attributes_t *attributes);
-psa_key_attributes_t shim_key_attributes_init(void);
 psa_status_t shim_get_key_attributes(psa_key_handle_t key_handle, psa_key_attributes_t *attributes);
+size_t shim_get_key_bits(const psa_key_attributes_t *attributes);
+psa_key_id_t shim_get_key_id(const psa_key_attributes_t *attributes);
+psa_key_lifetime_t shim_get_key_lifetime(const psa_key_attributes_t *attributes);
+psa_key_type_t shim_get_key_type(const psa_key_attributes_t *attributes);
+psa_key_usage_t shim_get_key_usage_flags(const psa_key_attributes_t *attributes);
+psa_key_attributes_t shim_key_attributes_init(void);
 
-void shim_set_key_algorithm(psa_key_attributes_t *attributes,
-                            psa_algorithm_t alg);
-
+void shim_set_key_algorithm(psa_key_attributes_t *attributes, psa_algorithm_t alg);
 void shim_set_key_bits(psa_key_attributes_t *attributes, size_t bits);
-
 void shim_set_key_id(psa_key_attributes_t *attributes, psa_key_id_t id);
-
-void shim_set_key_lifetime(psa_key_attributes_t *attributes,
-                           psa_key_lifetime_t lifetime);
-
+void shim_set_key_lifetime(psa_key_attributes_t *attributes, psa_key_lifetime_t lifetime);
 void shim_set_key_type(psa_key_attributes_t *attributes, psa_key_type_t type_);
-
-void shim_set_key_usage_flags(psa_key_attributes_t *attributes,
-                              psa_key_usage_t usage_flags);
+void shim_set_key_usage_flags(psa_key_attributes_t *attributes, psa_key_usage_t usage_flags);
 
 int shim_PSA_ALG_IS_HASH(psa_algorithm_t alg);
 int shim_PSA_ALG_IS_MAC(psa_algorithm_t alg);
--- a/psa-crypto-sys/src/c/shim.h
+++ b/psa-crypto-sys/src/c/shim.h
@@ -83,6 +83,7 @@ psa_algorithm_t shim_get_key_algorithm(const psa_key_attributes_t *attributes);
 psa_key_usage_t
 shim_get_key_usage_flags(const psa_key_attributes_t *attributes);
 psa_key_attributes_t shim_key_attributes_init(void);
+psa_status_t shim_get_key_attributes(psa_key_handle_t key_handle, psa_key_attributes_t *attributes);
 
 void shim_set_key_algorithm(psa_key_attributes_t *attributes,
                             psa_algorithm_t alg);
--- a/psa-crypto-sys/src/constants.rs
+++ b/psa-crypto-sys/src/constants.rs
@@ -90,6 +90,8 @@ pub const PSA_KEY_USAGE_DECRYPT: psa_key_usage_t = 512;
 pub const PSA_KEY_USAGE_SIGN: psa_key_usage_t = 1024;
 pub const PSA_KEY_USAGE_VERIFY: psa_key_usage_t = 2048;
 pub const PSA_KEY_USAGE_DERIVE: psa_key_usage_t = 4096;
+pub const PSA_KEY_SLOT_COUNT: isize = 32;
+pub const PSA_MAX_PERSISTENT_KEY_IDENTIFIER: psa_key_id_t = 0x3fff_ffff;
 
 #[cfg(feature = "implementation-defined")]
 pub const PSA_DRV_SE_HAL_VERSION: u32 = 5;
--- a/psa-crypto-sys/src/constants.rs
+++ b/psa-crypto-sys/src/constants.rs
@@ -90,8 +90,6 @@ pub const PSA_KEY_USAGE_DECRYPT: psa_key_usage_t = 512;
 pub const PSA_KEY_USAGE_SIGN: psa_key_usage_t = 1024;
 pub const PSA_KEY_USAGE_VERIFY: psa_key_usage_t = 2048;
 pub const PSA_KEY_USAGE_DERIVE: psa_key_usage_t = 4096;
-pub const PSA_KEY_SLOT_COUNT: isize = 32;
-pub const PSA_MAX_PERSISTENT_KEY_IDENTIFIER: psa_key_id_t = 0x3fff_ffff;
 
 #[cfg(feature = "implementation-defined")]
 pub const PSA_DRV_SE_HAL_VERSION: u32 = 5;
--- a/psa-crypto-sys/src/constants.rs
+++ b/psa-crypto-sys/src/constants.rs
@@ -10,15 +10,17 @@ use super::types::*;
 // PSA error codes
 pub const PSA_SUCCESS: psa_status_t = 0;
 pub const PSA_ERROR_GENERIC_ERROR: psa_status_t = -132;
-pub const PSA_ERROR_NOT_SUPPORTED: psa_status_t = -134;
 pub const PSA_ERROR_NOT_PERMITTED: psa_status_t = -133;
+pub const PSA_ERROR_NOT_SUPPORTED: psa_status_t = -134;
+pub const PSA_ERROR_INVALID_ARGUMENT: psa_status_t = -135;
+pub const PSA_ERROR_INVALID_HANDLE: psa_status_t = -136;
+pub const PSA_ERROR_BAD_STATE: psa_status_t = -137;
 pub const PSA_ERROR_BUFFER_TOO_SMALL: psa_status_t = -138;
 pub const PSA_ERROR_ALREADY_EXISTS: psa_status_t = -139;
 pub const PSA_ERROR_DOES_NOT_EXIST: psa_status_t = -140;
-pub const PSA_ERROR_BAD_STATE: psa_status_t = -137;
-pub const PSA_ERROR_INVALID_ARGUMENT: psa_status_t = -135;
 pub const PSA_ERROR_INSUFFICIENT_MEMORY: psa_status_t = -141;
 pub const PSA_ERROR_INSUFFICIENT_STORAGE: psa_status_t = -142;
+pub const PSA_ERROR_INSUFFICIENT_DATA: psa_status_t = -143;
 pub const PSA_ERROR_COMMUNICATION_FAILURE: psa_status_t = -145;
 pub const PSA_ERROR_STORAGE_FAILURE: psa_status_t = -146;
 pub const PSA_ERROR_HARDWARE_FAILURE: psa_status_t = -147;
@@ -28,8 +30,6 @@ pub const PSA_ERROR_INVALID_PADDING: psa_status_t = -150;
 pub const PSA_ERROR_CORRUPTION_DETECTED: psa_status_t = -151;
 pub const PSA_ERROR_DATA_CORRUPT: psa_status_t = -152;
 pub const PSA_ERROR_DATA_INVALID: psa_status_t = -153;
-pub const PSA_ERROR_INSUFFICIENT_DATA: psa_status_t = -143;
-pub const PSA_ERROR_INVALID_HANDLE: psa_status_t = -136;
 
 pub const PSA_MAX_KEY_BITS: usize = 65528;
 pub const PSA_KEY_TYPE_NONE: psa_key_type_t = 0;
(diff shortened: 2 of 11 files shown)


psa-crypto-sys/src/constants.rs
```diff
@@ -87,6 +90,8 @@ pub const PSA_KEY_USAGE_DECRYPT: psa_key_usage_t = 512;
 pub const PSA_KEY_USAGE_SIGN: psa_key_usage_t = 1024;
 pub const PSA_KEY_USAGE_VERIFY: psa_key_usage_t = 2048;
 pub const PSA_KEY_USAGE_DERIVE: psa_key_usage_t = 4096;
+pub const PSA_KEY_SLOT_COUNT: isize = 32;
+pub const PSA_MAX_PERSISTENT_KEY_IDENTIFIER: psa_key_id_t = 0x3fff_ffff;
```
hug-dev, line 94: These values seem to be Mbed Crypto specific. Ideally this crate is common to all PSA Crypto implementation so it would be better, I think, to declare those values in the code that needs them directly.

psa-crypto-sys/src/lib.rs
```diff
@@ -39,9 +39,9 @@ pub use types::*;
 
 #[cfg(feature = "implementation-defined")]
 pub use psa_crypto_binding::{
-    psa_close_key, psa_crypto_init, psa_destroy_key, psa_export_public_key, psa_generate_key,
-    psa_import_key, psa_key_attributes_t, psa_open_key, psa_reset_key_attributes, psa_sign_hash,
-    psa_verify_hash,
+    mbedtls_psa_crypto_free, psa_close_key, psa_crypto_init, psa_destroy_key,
```
hug-dev, line 42: Same for `mbedtls_psa_crypto_free`, I don't think we should declare it here.

psa-crypto/src/operations/key_management.rs
```diff
@@ -227,17 +219,87 @@ pub fn export_public(key: Id, data: &mut [u8]) -> Result<usize> {
...
+/// # };
+/// psa_crypto::init().unwrap();
+/// let my_key = key_management::generate(attributes, None).unwrap();
+/// //...
+/// let key_attributes = key_management::get_key_attributes(my_key);
+/// ```
+
+pub fn get_key_attributes(key: Id) -> Result<Attributes> {
+    initialized()?;
+    let mut key_attributes = unsafe { psa_crypto_sys::psa_key_attributes_init() };
```
hug-dev, line 271: I think if we follow the specs, the attributes created here might allocate some extra space that needs to be freed with a call to `psa_reset_key_attributes` (`Attributes::reset`). I think that might be needed on every path leaving the function once `psa_key_attributes_init` has been called.

psa-crypto/src/operations/key_management.rs
```diff
@@ -227,17 +219,87 @@ pub fn export_public(key: Id, data: &mut [u8]) -> Result<usize> {
...
+fn complete_new_key_operation(
+    key_lifetime: Lifetime,
+    id: psa_key_id_t,
+    handle: psa_key_handle_t,
+    operation_result: Result<()>,
+) -> Result<Id> {
+    if key_lifetime != Lifetime::Volatile {
+        Status::from(unsafe { psa_crypto_sys::psa_close_key(handle) }).to_result()?;
+    }
+    operation_result?;
```
hug-dev, line 296: Do you think that if `psa_generate_key` or `psa_import_key`  failed you still need to close the key? I would have thought that if those operations fail, they do not open the key.
The documentation of `psa_generate_key` says:
```
 * \param[out] handle       On success, a handle to the newly created key.
 *                          \c 0 on failure.
```
So I think this is fine not passing `operation_result` to this function (checking in the caller if it failed or not).

psa-crypto/src/types/key.rs
```diff
@@ -12,6 +12,9 @@ use crate::types::status::{Error, Result};
 #[cfg(feature = "with-mbed-crypto")]
 use core::convert::{TryFrom, TryInto};
 use log::error;
+pub use psa_crypto_sys::{
+    psa_key_id_t as key_id_type, PSA_KEY_SLOT_COUNT, PSA_MAX_PERSISTENT_KEY_IDENTIFIER,
```
hug-dev, line 16: Is the renaming of `psa_key_id_t` required? I think it is easier to understand when you have the original type name but that might be a personal opinion 😋

psa-crypto/src/types/status.rs
```diff
@@ -218,19 +221,20 @@ impl From<Error> for psa_crypto_sys::psa_status_t {
...
             Error::InvalidPadding => psa_crypto_sys::PSA_ERROR_INVALID_PADDING,
             Error::InsufficientData => psa_crypto_sys::PSA_ERROR_INSUFFICIENT_DATA,
             Error::InvalidHandle => psa_crypto_sys::PSA_ERROR_INVALID_HANDLE,
-            e => {
+            // Commented out as all errors are currently matched against and this causes compilation error
+            /*e => {
                 error!("No equivalent of {:?} to a psa_status_t.", e);
                 psa_crypto_sys::PSA_ERROR_GENERIC_ERROR
-            }
+            }*/
```
hug-dev, line 237: That's great, thanks for completing this conversion. Feel free to delete that block altogether 👍

psa-crypto/tests/mod.rs
```diff
@@ -0,0 +1,113 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+use psa_crypto::types::algorithm::{Algorithm, AsymmetricSignature, Hash};
+use psa_crypto::types::key::{Attributes, Lifetime, Policy, Type, UsageFlags};
+
+#[test]
+fn generate_integration_test() {
```
hug-dev, line 8: Nice, thanks for the new integration tests!
Nice to see that the CI passed them as well!

psa-crypto/src/operations/asym_signature.rs
```diff
@@ -75,10 +75,9 @@ pub fn sign_hash(
             &mut signature_length,
         )
     })
-    .to_result()?;
-
+    .to_result();
     key.close_handle(handle)?;
-
+    sign_res?;
```
hug-dev, line 80: That is actually quite clever 😄 Never thought of doing it like that before.

psa-crypto/src/operations/key_management.rs
```diff
@@ -227,17 +219,87 @@ pub fn export_public(key: Id, data: &mut [u8]) -> Result<usize> {
...
+/// #         }.into(),
+/// #     },
+/// # };
+/// psa_crypto::init().unwrap();
+/// let my_key = key_management::generate(attributes, None).unwrap();
+/// //...
+/// let key_attributes = key_management::get_key_attributes(my_key);
+/// ```
+
+pub fn get_key_attributes(key: Id) -> Result<Attributes> {
```
hug-dev, line 269: In a way, this function is a contructor of `Attributes` so I think it could be placed inside the `impl Attributes` block. In that case, we could rename it, using namespacing. What about something like `from_key`? It would then be `Attributes::from_key(...)` to invoke it.
Agree that we would deviate from the PSA Crypto naming but I think it is fine if we make it more Rust-like!

psa-crypto/src/operations/asym_signature.rs
ionut-arm, line 80: For these cases where you *need* to do a cleanup operation that may fail (after some other core operation might've failed in the first place) we have to decide: if both fail, which error do we want to be returned? I'd say the error on the "core" operation should take precedence, in which case the two branches (as in, if the core operation failed or not) will be different

ionut-arm: Also, I wish we had a better approach to this kind of problem where you have some cleanup to do whenever you exit the function. In this case it's fine since there's not much branching, but in some cases it might end up being a pain having to call that `close_handle` for each possible path before doing `?`.

psa-crypto/src/operations/key_management.rs
ionut-arm: `from_key_id` - `Id` is not a key per se and we keep naming things as if it is.

psa-crypto/src/operations/key_management.rs
ionut-arm: Hence my pet peeve with these cleanup methods.
What happens if both `close_handle` and `psa_reset_key_attributes` fail, which value do we return? Maybe we should really return a generic error when some of these "internal" methods fail, that the user should not really know about. 

An alternative would be to make a "layered" function, where for each variable that needs cleaning up we wrap everything after it in a function that returns a `Result`. If the result is `Ok`, attempt to clean up and if that fails return an error, otherwise return the `Ok` you got. If what you got was `Err`, then do the clean up anyway and return the `Err` you got from the next layer function. If you have two things that you need to clean, unfortunately you'd have 2 "layer" functions. But the benefit is that you're sure that you clean up everything and you return the `Err` that's most appropriate.

ionut-arm:
> 
> 
> Thank you for the changes! 👍

Disregard the approval for now, my mistake 😅

psa-crypto/src/operations/asym_signature.rs
sbailey-arm: I originally had it returning the first operation error, then the close handle error but after a brief discussion with @hug-dev, it was changed to how it is now. I think the reason was the [PSA API states keys aren't used](https://armmbed.github.io/mbed-crypto/html/appendix/history.html#changes-between-1-0-beta-3-and-1-0-0?highlight=psa_key_handle_t), IDs are, so it should be getting removed at some point (soon? 😄) anyway.

psa-crypto-sys/src/constants.rs
sbailey-arm: Yes, they are Mbed Crypto specific. Should they be moved to parsec?

psa-crypto/src/types/key.rs
sbailey-arm: No, not required. I think I just did it so it was clear in my head when I was making the change, so I didn't trip up with the other reference to the same type from the binding, when I was swapping them over. I agree, it'll be clearer if the original name is kept 😃

psa-crypto/src/types/status.rs
sbailey-arm: Will remove

psa-crypto/src/operations/asym_signature.rs
ionut-arm: Yes, I think that makes sense for this case, my comment was more general - we should have a statement somewhere as to how to prioritize these errors throughout the library. I'd even say that this applies to Parsec too. The last error to occur is not necessarily the one we should return.

psa-crypto-sys/src/constants.rs
hug-dev: Yes I think it will be better.

psa-crypto/src/operations/key_management.rs
sbailey-arm: I have checked `psa_generate_key` and `psa_import_key` - you are correct, a handle should not be created unless the operation succeeds.

psa-crypto-sys/src/lib.rs
sbailey-arm: I'm not sure how we should deal with this. For `psa-crypto` to be portable to any PSA Crypto implementation, we can't have this `mbedtls` specific method declared here, but I'm not sure how we remove it from `psa-crypto` and allow Parsec to call it without binding straight through to `mbedtls`.

hug-dev: I think this is is used inside Mbed Crypto for testing but should not be required in this crate. If it is, and creating leaks in our applications, we should probably discuss this at a PSA Crypto level.

psa-crypto-sys/src/c/shim.h
```diff
@@ -83,6 +83,7 @@ psa_algorithm_t shim_get_key_algorithm(const psa_key_attributes_t *attributes);
 psa_key_usage_t
 shim_get_key_usage_flags(const psa_key_attributes_t *attributes);
 psa_key_attributes_t shim_key_attributes_init(void);
+psa_status_t shim_get_key_attributes(psa_key_handle_t key_handle, psa_key_attributes_t *attributes);
```
egrimley-arm, line 86: It would be nice to tidy this code up a bit while you're at it. How about having the return type consistently on the same line as the function name and putting the functions in alphabetical order as they already mostly are?

psa-crypto-sys/src/constants.rs
```diff
@@ -25,6 +25,9 @@ pub const PSA_ERROR_HARDWARE_FAILURE: psa_status_t = -147;
 pub const PSA_ERROR_INSUFFICIENT_ENTROPY: psa_status_t = -148;
 pub const PSA_ERROR_INVALID_SIGNATURE: psa_status_t = -149;
 pub const PSA_ERROR_INVALID_PADDING: psa_status_t = -150;
+pub const PSA_ERROR_CORRUPTION_DETECTED: psa_status_t = -151;
+pub const PSA_ERROR_DATA_CORRUPT: psa_status_t = -152;
+pub const PSA_ERROR_DATA_INVALID: psa_status_t = -153;
 pub const PSA_ERROR_INSUFFICIENT_DATA: psa_status_t = -143;
```
egrimley-arm, line 31: The last two (-143 and -136) seem to be in the wrong place but that's not part of this patch so I shouldn't really complain.

egrimley-arm: Remind me why these values can't come from the C library somehow? In a previous patch we had something like `pub const PSA_SOMETHING: psa_something_t = shim_PSA_SOMETHING;` in the Rust code and `const psa_something_t shim_PSA_SOMETHING = PSA_SOMETHING;` in `shim.h`. It would be nice to explain in a comment at the top of `constants.rs` why these values are being repeated like this.

psa-crypto-sys/src/constants.rs
egrimley-arm: Presumably `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` will be replaced with something else in PSA Crypto 1.0?

psa-crypto/src/operations/key_management.rs
egrimley-arm: I think `psa_reset_key_attributes` can't fail, though that doesn't invalidate the general point, just that particular example!

egrimley-arm: (Also, I think you don't need to call `psa_reset_key_attributes` if calling `psa_key_attributes_init` is all you've done with the `psa_key_attributes_t`, and according to the 1.0 spec there seem to be certain `psa_set_key_` functions that you can call without needing to do the reset.)

psa-crypto-sys/src/constants.rs
hug-dev: > Presumably PSA_MAX_PERSISTENT_KEY_IDENTIFIER will be replaced with something else in PSA Crypto 1.0?

That's a good point, on section 9.1.1, for `psa_key_id_t`, it defines ranges of key identifiers. For example for application key identifiers, it can range from `PSA_KEY_ID_USER_MIN` (`0x00000001`) to `PSA_KEY_ID_USER_MAX` (`0x3fffffff`). Which is the same value as the one above 😄! So I guess as `PSA_KEY_SLOT_COUNT` should be removed, `PSA_MAX_PERSISTENT_KEY_IDENTIFIER` can be renamed to `PSA_KEY_ID_USER_MAX`

psa-crypto-sys/src/constants.rs
hug-dev: We removed from the shim all the values that are defined by the specification. That way, it allows you to use those value, without having to link with a PSA Crypto implementation (that is the `implementation-defined` feature). That is useful for Parsec interface where we re-use some of these values but do not want to link to a PSA Crypto implementation.

psa-crypto-sys/src/c/shim.h
ionut-arm: Is it possible to add CI linting (potentially `clang-format`)? Wouldn't take care of the alphabetical order, but it would keep other things tidy. Might be a bit overkill, tho

psa-crypto-sys/src/constants.rs
ionut-arm: At least I hope I removed all the duplication, as I did everything manually

psa-crypto/src/lib.rs
```diff
@@ -43,6 +43,7 @@
 
 #[cfg(feature = "with-mbed-crypto")]
 pub mod operations;
+#[cfg(feature = "with-mbed-crypto")]
```
hug-dev, line 46: I don't think this `cfg` gate is needed for that module!

psa-crypto/src/types/key.rs
```diff
@@ -255,6 +257,53 @@ impl Attributes {
...
+    /// #             hash_alg: Hash::Sha256.into(),
+    /// #         }.into(),
+    /// #     },
+    /// # };
+    /// psa_crypto::init().unwrap();
+    /// let my_key_id = key_management::generate(attributes, None).unwrap();
+    /// //...
+    /// let key_attributes = Attributes::from_key_id(my_key_id);
+    /// ```
+    pub fn from_key_id(key_id: Id) -> Result<Self> {
```
ionut-arm, line 293: Would it be worth having a test for this?

psa-crypto-sys/src/constants.rs
egrimley-arm: I hadn't realised some values are defined by the specification. Should that be mentioned in a comment in `constants.rs`?

ionut-arm: I think that makes sense - we do have a module comment in this file, it might be worth expanding that to explain the logic of what's hidden by `implementation-defined` - there's a brief explanation [here](https://github.com/parallaxsecond/rust-psa-crypto/tree/master/psa-crypto-sys) but it should really be in docs.rs too.

psa-crypto/src/lib.rs
sbailey-arm: For `types`? Now `get_key_attributes` is in `Attributes`, it needs access to some `mbed-crypto` specific code. Without that, `ci.sh` gives a load of `not found in` errors

sbailey-arm: Actually, I could make it more local. Adding a gate for `use crate::initialized` and one for the method.

psa-crypto/src/types/key.rs
hug-dev: As there is an example, let's say it is fine for now but think about adding one in the future 🤡

PR #31, opened 2020-06-22
Added macro calls for sign output size and export key buffer size

sbailey-arm:

### DIFF
--- a/psa-crypto-sys/src/shim_methods.rs
+++ b/psa-crypto-sys/src/shim_methods.rs
@@ -171,3 +171,19 @@ pub fn PSA_KEY_TYPE_DH_KEY_PAIR(group: psa_dh_group_t) -> psa_key_type_t {
 pub fn PSA_KEY_TYPE_DH_PUBLIC_KEY(group: psa_dh_group_t) -> psa_key_type_t {
     unsafe { psa_crypto_binding::shim_PSA_KEY_TYPE_DH_PUBLIC_KEY(group) }
 }
+
+pub unsafe fn PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(key_type: psa_key_type_t) -> psa_key_type_t {
+    psa_crypto_binding::shim_PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(key_type)
+}
+
+pub unsafe fn PSA_SIGN_OUTPUT_SIZE(
+    key_type: psa_key_type_t,
+    key_bits: usize,
+    alg: psa_algorithm_t,
+) -> usize {
+    psa_crypto_binding::shim_PSA_SIGN_OUTPUT_SIZE(key_type, key_bits, alg)
+}
+
+pub unsafe fn PSA_EXPORT_KEY_OUTPUT_SIZE(key_type: psa_key_type_t, key_bits: usize) -> usize {
+    psa_crypto_binding::shim_PSA_KEY_EXPORT_MAX_SIZE(key_type, key_bits)
+}
--- a/psa-crypto/src/types/key.rs
+++ b/psa-crypto/src/types/key.rs
@@ -6,6 +6,8 @@
 #![allow(deprecated)]
 #[cfg(feature = "with-mbed-crypto")]
 use crate::initialized;
+#[cfg(feature = "with-mbed-crypto")]
+use crate::types::algorithm::AsymmetricSignature;
 use crate::types::algorithm::{Algorithm, Cipher};
 #[cfg(feature = "with-mbed-crypto")]
 use crate::types::status::Status;
@@ -305,6 +307,50 @@ impl Attributes {
         get_attributes_res?;
         Ok(attributes?)
     }
+
+    /// Sufficient size for a buffer to export the key, if supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn export_key_output_size(self) -> Result<usize> {
+        Attributes::export_key_output_size_base(self.key_type, self.bits)
+    }
+
+    /// Sufficient size for a buffer to export the public key, if supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn export_public_key_output_size(self) -> Result<usize> {
+        match self.key_type {
+            Type::RsaKeyPair
+            | Type::RsaPublicKey
+            | Type::EccKeyPair { .. }
+            | Type::EccPublicKey { .. }
+            | Type::DhKeyPair { .. }
+            | Type::DhPublicKey { .. } => {
+                let pub_type = self.key_type.key_type_public_key_of_key_pair()?;
+                Attributes::export_key_output_size_base(pub_type, self.bits)
+            }
+            _ => Err(Error::InvalidArgument),
+        }
+    }
+
+    /// Sufficient size for a buffer to export the given key type, if supported
+    #[cfg(feature = "with-mbed-crypto")]
+    fn export_key_output_size_base(key_type: Type, bits: usize) -> Result<usize> {
+        let size =
+            unsafe { psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(key_type.try_into()?, bits) };
+        if size > 0 {
+            Ok(size)
+        } else {
+            Err(Error::NotSupported)
+        }
+    }
+
+    /// Sufficient buffer size for a signature using the given key, if the key is supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn sign_output_size(self, alg: AsymmetricSignature) -> Result<usize> {
+        self.compatible_with_alg(Algorithm::AsymmetricSignature(alg))?;
+        Ok(unsafe {
+            psa_crypto_sys::PSA_SIGN_OUTPUT_SIZE(self.key_type.try_into()?, self.bits, alg.into())
+        })
+    }
 }
 
 /// The lifetime of a key indicates where it is stored and which application and system actions
@@ -405,6 +451,27 @@ impl Type {
         match self {
             Type::DhKeyPair { .. } => true,
             _ => false,
+        }
+    }
+
+    /// If key is public or key pair, returns the corresponding public key type.
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn key_type_public_key_of_key_pair(self) -> Result<Type> {
+        match self {
+            Type::RsaKeyPair
+            | Type::RsaPublicKey
+            | Type::EccKeyPair { .. }
+            | Type::EccPublicKey { .. }
+            | Type::DhKeyPair { .. }
+            | Type::DhPublicKey { .. } => {
+                Ok(
+                    unsafe {
+                        psa_crypto_sys::PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(self.try_into()?)
+                    }
+                    .try_into()?,
+                )
+            }
+            _ => Err(Error::InvalidArgument),
         }
     }
 }
--- a/ci.sh
+++ b/ci.sh
@@ -31,8 +31,10 @@ fi
 #############
 # Run tests #
 #############
-RUST_BACKTRACE=1 cargo test
+RUST_BACKTRACE=1 cargo test -- --test-threads=1
 
+# Remove mbedtls directory if it exists
+rm -rf psa-crypto/mbedtls
 ################################
 # Check feature configurations #
 ################################
--- a/psa-crypto-sys/src/c/shim.c
+++ b/psa-crypto-sys/src/c/shim.c
@@ -236,3 +236,21 @@ shim_PSA_KEY_TYPE_DH_PUBLIC_KEY(psa_dh_group_t group)
 {
 	return PSA_KEY_TYPE_DH_PUBLIC_KEY(group);
 }
+
+psa_key_type_t
+shim_PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(psa_key_type_t key_type)
+{
+    return PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(key_type);
+}
+
+size_t
+shim_PSA_SIGN_OUTPUT_SIZE(psa_key_type_t key_type, size_t key_bits, psa_algorithm_t alg)
+{
+    return PSA_SIGN_OUTPUT_SIZE(key_type, key_bits, alg);
+}
+
+size_t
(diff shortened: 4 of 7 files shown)


psa-crypto/src/types/key.rs
```diff
@@ -305,6 +305,37 @@ impl Attributes {
...
+                self.key_type.try_into()?,
+                self.bits,
+            )),
+            _ => Err(Error::NotSupported),
+        }
+    }
+
+    /// Sufficient buffer size for a signature using the given key, if the key is supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn sign_output_size(self) -> Result<usize> {
```
hug-dev, line 326: I think the `alg` parameter of `PSA_SIGN_OUTPUT_SIZE` macro should not be the same as the `permitted_algorithms` field of the attributes but should come from as another `alg` parameter of this function.
My understanding is that the `permitted_algorithms` field only specify the authorised algorithms (there can be multiple) that these attributes allow as opposed as the specific algorithm to use for a particular operation.
Even if the results will be the same, I think it would be better this way, to add an `alg` parameter (of type `AsymmetricSignature`) to this function!

psa-crypto-sys/src/shim_methods.rs
```diff
@@ -171,3 +171,15 @@ pub fn PSA_KEY_TYPE_DH_KEY_PAIR(group: psa_dh_group_t) -> psa_key_type_t {
 pub fn PSA_KEY_TYPE_DH_PUBLIC_KEY(group: psa_dh_group_t) -> psa_key_type_t {
     unsafe { psa_crypto_binding::shim_PSA_KEY_TYPE_DH_PUBLIC_KEY(group) }
 }
+
+pub fn PSA_SIGN_OUTPUT_SIZE(
```
hug-dev, line 175: I am thinking if those macros should be safe or not... The specification says:

> If the parameters are not valid, the return value is unspecified.

and I am wondering if `unspecified` is also `unsafe` 😋 Maybe it is safer to add `unsafe` here, ironically

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +305,37 @@ impl Attributes {
...
+        match self.key_type {
```
hug-dev, line 327: I think here you can re-use the [`compatible_with_alg`](https://docs.rs/psa-crypto/0.2.0/psa_crypto/types/key/struct.Attributes.html#method.compatible_with_alg) method of `Attributes` which implements similar logic and do something like:
```rust
self.compatible_with_alg(alg)?;
Ok(PSA_SIGN_OUTPUT_SIZE(..))
```

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +305,37 @@ impl Attributes {
...
+    pub fn export_key_output_size(self) -> Result<usize> {
+        match self.key_type {
+            Type::RsaKeyPair
+            | Type::RsaPublicKey
+            | Type::EccKeyPair { .. }
+            | Type::EccPublicKey { .. } => Ok(psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(
+                self.key_type.try_into()?,
+                self.bits,
+            )),
+            _ => Err(Error::NotSupported),
```
hug-dev, line 320: Maybe `Error:InvalidArgument` is better here to follow the specs.

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +305,37 @@ impl Attributes {
         get_attributes_res?;
         Ok(attributes?)
     }
+
+    /// Sufficient size for a buffer to export the key, if supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn export_key_output_size(self) -> Result<usize> {
+        match self.key_type {
```
hug-dev, line 312: The specification says that this macro allows for `key_type`:

> `key_type` A public key or key pair key type.

So I think you can add DH both for key pair and public key as a valid type as well.

psa-crypto-sys/src/shim_methods.rs
```diff
@@ -171,3 +171,15 @@ pub fn PSA_KEY_TYPE_DH_KEY_PAIR(group: psa_dh_group_t) -> psa_key_type_t {
...
+    key_type: psa_key_type_t,
+    key_bits: usize,
+    alg: psa_algorithm_t,
+) -> usize {
+    unsafe { psa_crypto_binding::shim_PSA_SIGN_OUTPUT_SIZE(key_type, key_bits, alg) }
+}
+
+pub fn PSA_EXPORT_KEY_OUTPUT_SIZE(key_type: psa_key_type_t, key_bits: usize) -> usize {
```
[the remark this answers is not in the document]
sbailey-arm: I've gone off the [API changelog](https://armmbed.github.io/mbed-crypto/html/appendix/history.html?highlight=psa_key_export_max_size) (scroll down to the highlighted bit) which states `PSA_KEY_EXPORT_MAX_SIZE() → PSA_EXPORT_KEY_OUTPUT_SIZE()`. Though the old macro is Parsec did specifically state `public`. Not sure why there is an inconsistency?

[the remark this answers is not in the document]
hug-dev: Ah you are right, my bad! Seems like `PSA_EXPORT_KEY_OUTPUT_SIZE` works for both public keys and key pairs. I think in Parsec we only support `psa_export_public_key` in which case we will need a way to get the output size of the public part of a key pair (to have the size only of the public part and not the whole key).

psa-crypto/src/types/key.rs
sbailey-arm: I was wondering about that, there seemed to be no specific algorithm for a given Attributes struct. Looking at the way it used to work, it did use `permitted_algorithm`,  just like I've done (it converted `Attributes` into `psa_key_attributes_t` which does `attributes.policy.permitted_algorithms.try_into()?` to get the algorithm). 

Can we not fish out the `AsymmetricSignature` from self and if it doesn't have one, return the error as it does currently?
Something like:
```rust
if let Algorithm::AsymmetricSignature(sign_alg) = self.policy.permitted_algorithms {
    Ok(psa_crypto_sys::PSA_SIGN_OUTPUT_SIZE(
        self.key_type.try_into()?,
        self.bits,
        sign_alg.into(),
    ))
}
else {
    Err(Error::InvalidArgument)
}
```

psa-crypto-sys/src/shim_methods.rs
sbailey-arm: Can't we do enough checks to make it safe? As in, if we have the checks in place to ensure we only ever pass the macros valid arguments, it shouldn't ever return an unspecified value.

psa-crypto/src/types/key.rs
sbailey-arm: Is DH the same as DSA? I've been looking at [this](https://github.com/ARMmbed/mbedtls/blob/bd004e862d8479586d7622811d02c65f44544125/include/psa/crypto_sizes.h#L652) to try and make sure its not possible to pass it something invalid.

ionut-arm: > Is DH the same as DSA?

No, Diffie-Hellman is a type of key exchange, Digital Signature Algorithm is (as its name says) an asymmetric signature algorithm. 

For DH public keys, the spec dictates:
> For Diffie-Hellman key exchange public keys, with key types for which PSA_KEY_TYPE_IS_DH_PUBLIC_KEY is true, the format is the representation of the public key y = g^x mod p as a big-endian byte string. The length of the byte string is the length of the base prime p in bytes.

Which brings me to another realisation - that method there, `export_key_output_size` is improperly named, as it refers only to public keys, not to keys in general.

psa-crypto/src/types/key.rs
hug-dev: I think that before we were doing things wrong, mixing and using the policy algorithm as the algorithm used for any crypto operation. We [changed that and added an alg field](https://github.com/parallaxsecond/parsec-operations/commit/d301c62d88c10db7501f2eed8bcba17ab0c24202) to specifically split those two semantic units which are the policy algorithms (the algorithms permitted to be used with a key, they can be multiple because you can have `SignHash::Any` as an hash algorithm) and the specific algorithm used for a crypto operation. This last one, [can not be a policy algorithm](https://armmbed.github.io/mbed-crypto/html/api/ops/sign.html#c.PSA_ALG_ANY_HASH).

Following the same logic, I think it would be better splitting things here as well, as in not using the policy algorithm as the one to use to check the output size.

psa-crypto-sys/src/shim_methods.rs
hug-dev: > Can't we do enough checks to make it safe?

I definitely agree and that is what we are trying to do in the `psa-crypto` crate. The `psa-crypto-sys` should only be the thin Rust wrapper to allow calling the C API and can be unsafe.

psa-crypto/src/types/key.rs
ionut-arm: Ah, I see, you're using the spec-provided macros, I think that's alright (although, will those disappear in the future?). But we do have to make sure we document whether this method refers to the full key or just the public part

psa-crypto-sys/src/shim_methods.rs
sbailey-arm: Ah ok, I see what you mean now. Yes I think they should be marked as unsafe as you suggested in `psa-crypto-sys`.

psa-crypto/src/types/key.rs
sbailey-arm: Ah ok. So where should the algorithm come from? Should there be another `Algorithm` attribute in the struct or is it held separately?

hug-dev: I think it should be part of the method's signature!

```rust
pub fn sign_output_size(self, alg: AsymmetricSignature) -> Result<usize>
```

sbailey-arm: Yeah, I meant where does the caller get the value from :laughing:. I've got it though, `op.alg`.

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +305,37 @@ impl Attributes {
...
+    }
+
+    /// Sufficient buffer size for a signature using the given key, if the key is supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn sign_output_size(self) -> Result<usize> {
+        match self.key_type {
+            Type::RsaPublicKey
+            | Type::RsaKeyPair
+            | Type::EccPublicKey { .. }
+            | Type::EccKeyPair { .. } => Ok(psa_crypto_sys::PSA_SIGN_OUTPUT_SIZE(
```
ionut-arm, line 331: > If the parameters are a valid combination that is not supported by the implementation, this macro must return either a sensible size or 0.

Do we want to return that 0 or do we want to convert that case into an error? Surely a size of 0 is not helpful to anyone

hug-dev: You mean checking the return value and return like `NotSupported` if `0`?

ionut-arm: Yeah, something like that - but I don't know if that contradicts the spec. Tbh we're returning a `Result` anyway so it's not a big deviation

psa-crypto-sys/src/shim_methods.rs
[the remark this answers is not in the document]
sbailey-arm: Currently I don't think `mbedtls` supports `PSA_EXPORT_PUBLIC_KEY_OUTPUT_SIZE` (at least I can't find it using Githubs search feature). So does that mean we should add `psa_export_public_key_output_size and do some jiggery pokery with `PSA_EXPORT_KEY_OUTPUT_SIZE`? That way we implement the spec in `psa-crypto` and Parsec can use it. Or do we have to ensure that Parsec only supplies the public key to `export_key_output_size`?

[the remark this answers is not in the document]
hug-dev: Could we do something like (as written in the spec): `PSA_EXPORT_KEY_OUTPUT_SIZE(PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR(key_type), key_bits)`? If `PSA_KEY_TYPE_PUBLIC_KEY_OF_KEY_PAIR` is available?

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +307,50 @@ impl Attributes {
...
+            Ok(size)
+        } else {
+            Err(Error::NotSupported)
+        }
+    }
+
+    /// Sufficient buffer size for a signature using the given key, if the key is supported
+    #[cfg(feature = "with-mbed-crypto")]
+    pub fn sign_output_size(self, alg: AsymmetricSignature) -> Result<usize> {
+        self.compatible_with_alg(Algorithm::AsymmetricSignature(alg))?;
```
hug-dev, line 349: This is perfectly fine like this (no need to change), just showing for reference that you can also do:
```rust
self.compatible_with_alg(alg.into())?;
```

psa-crypto/src/types/key.rs
```diff
@@ -305,6 +307,50 @@ impl Attributes {
...
+    /// Sufficient size for a buffer to export the given key type, if supported
+    #[cfg(feature = "with-mbed-crypto")]
+    fn export_key_output_size_base(key_type: Type, bits: usize) -> Result<usize> {
+        let size =
+            unsafe { psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(key_type.try_into()?, bits) };
+        if size > 0 {
+            Ok(size)
+        } else {
+            Err(Error::NotSupported)
+        }
```
ionut-arm, line 343: Same as what Hugues said, this works fine, no need to change, just for reference:
```rust
match unsafe { 
    psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(key_type.try_into()?, bits) 
  } {
    0 => Err(Error::NotSupported),
    val => Ok(val),
}
```

sbailey-arm: Ah yes, I was sure there would be a more succinct way of writing that bit, didn't think of doing it like your example!

hug-dev:
`thread 'generate_integration_test' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidHandle', psa-crypto/tests/mod.rs:97:23`

I am pretty sure Travis fails because of a data race in `psa_close_key` that is called simultaneously in our test. I propose to fix the problem in a later PR and force the tests to be single threaded for now.

You can apply the following change to `ci.sh`:

```diff
-RUST_BACKTRACE=1 cargo test
+RUST_BACKTRACE=1 cargo test -- --test-threads=1
```

PR #32, opened 2020-06-22
Add zeroize on drop for all types

ionut-arm:
This commit adds `zeroize` when dropping all PSA types declared in the
psa-crypto crate.

### DIFF
--- a/psa-crypto/src/types/algorithm.rs
+++ b/psa-crypto/src/types/algorithm.rs
@@ -12,11 +12,12 @@ use core::convert::{TryFrom, TryInto};
 #[cfg(feature = "with-mbed-crypto")]
 use log::error;
 use serde::{Deserialize, Serialize};
+use zeroize::Zeroize;
 
 /// Enumeration of possible algorithm definitions.
 /// Each variant of the enum contains a main algorithm type (which is required for
 /// that variant).
-#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Zeroize)]
 pub enum Algorithm {
     /// An invalid algorithm identifier value.
     /// `None` does not allow any cryptographic operation with the key. The key can still be
@@ -53,7 +54,7 @@ impl Algorithm {
     /// });
     /// assert!(hmac.is_hmac());
     /// ```
-    pub fn is_hmac(self) -> bool {
+    pub fn is_hmac(&self) -> bool {
         match self {
             Algorithm::Mac(mac_alg) => mac_alg.is_hmac(),
             _ => false,
@@ -62,7 +63,7 @@ impl Algorithm {
 }
 
 /// Enumeration of hash algorithms supported.
-#[derive(Copy, Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize, Zeroize)]
 #[allow(deprecated)]
 pub enum Hash {
     /// MD2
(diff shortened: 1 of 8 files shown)


psa-crypto/src/types/mod.rs
```diff
@@ -2,7 +2,44 @@
...
+zeroize_on_drop!(key::EccFamily);
+zeroize_on_drop!(key::Type);
+zeroize_on_drop!(key::Lifetime);
+zeroize_on_drop!(key::Policy);
+zeroize_on_drop!(key::Id);
+zeroize_on_drop!(key::Attributes);
+zeroize_on_drop!(key::UsageFlags);
+
+zeroize_on_drop!(status::Status);
+zeroize_on_drop!(status::Error);
```
hug-dev, line 45: Do all those types always carry confidential information? Do they all contain valuable data for an attacker?

psa-crypto/src/types/algorithm.rs
```diff
@@ -578,10 +579,10 @@ impl TryFrom<psa_crypto_sys::psa_algorithm_t> for Algorithm {
 impl TryFrom<Algorithm> for psa_crypto_sys::psa_algorithm_t {
     type Error = Error;
     fn try_from(alg: Algorithm) -> Result<Self> {
-        match alg {
+        match &alg {
```
hug-dev, line 582: Why is it necessary to match on the reference here and clone later instead of matching the value and passing ownership?

psa-crypto/src/types/mod.rs
ionut-arm: Don't think so, but I went for the full interpretation of the spec where all temporary values must be cleaned up.

psa-crypto/src/types/algorithm.rs
ionut-arm: Because for a type that implements `Drop` you can't take any elements out (otherwise you might not have everything you need in the object once you get into `drop`)

ionut-arm:
> 
> 
> Thanks for trying this out!
> A question in the review and another comment here:
> 
>     * if we go this way, the performance cost of zeroing on drop will be signicative for embedded targets I believe. We should make sure to minimize the amount of `clone()`-ing we do inside the stack. That could be through using references to the structures that are not `Copy` instead of passing them by values. I believe this is similar to the PSA Crypto API where the key attributes for example are often passed by reference to the different functions.

Yes, I've started working on that, I think most of those clones are removed (actually - all of them?), I'm only wondering whether we need to keep both conversions from `&T` to `psa_crypto_sys::T` and from `T` to the same?

psa-crypto/src/types/mod.rs
ionut-arm: Plus if you think about a full description of a key, you need to clear all the params associated with it - I chose the easy way out 😅 otherwise I'd have to implement `Zeroize` by hand on the types that don't have all inner values zeroize'able

ionut-arm: Also, if there's an efficiency problem here I can just make the `zeroize_on_drop` conditional on a feature.

hug-dev: Linking with [this issue](https://github.com/parallaxsecond/parsec/issues/122) for reference. I am not sure if the specification is that strict [on the section about memory cleanup](https://armmbed.github.io/mbed-crypto/html/overview/implementation.html#memory-cleanup).

> All temporary data used during the execution of a function, such as stack buffers, must be wiped before the function returns.

"stack buffer" given as an example might imply that this rule applies only to buffers.

> Implementations must wipe all sensitive data from memory when it is no longer used. It is recommended that they wipe this sensitive data as soon as possible.

Question is to know if those are parts of the "sensitive data". I would be careful about implementing such an impactful change if it is not needed, or negligible next to the memory cleaning of buffers. I will ask about that point.

ionut-arm: Sure, tbf it's probably a good idea to not have the memory wiping by default - I think having the `Zeroize` derived on the types makes sense, though

ionut-arm: We can also run some tests and see the exact performance figure with and without it

ionut-arm:
Dropping this, might reopen another with some of these changes

PR #36, opened 2020-07-03
Add an env var for static linking to Mbed Crypto

hug-dev:
This is useful when this crate is used deep in the dependency tree.


hug-dev:
The current use-case is for it to be used by the SE driver so I would have to update and release new versions for `psa-crypto-sys`, `psa-crypto`, `parsec-interface` and `parsec-client` 🤡

hug-dev:
Or I will probably just release a new version of `psa-crypto-sys` and `patch` it in the SE driver.

ionut-arm:
Feel free to update the other repo(s) however you like

PR #37, opened 2020-07-03
Added asymmetric encrypt and decrypt to psa-crypto and psa-crypto-sys

sbailey-arm:


psa-crypto-sys/src/shim_methods.rs
```diff
@@ -88,6 +88,10 @@ pub fn PSA_ALG_IS_ASYMMETRIC_ENCRYPTION(alg: psa_algorithm_t) -> bool {
     unsafe { psa_crypto_binding::shim_PSA_ALG_IS_ASYMMETRIC_ENCRYPTION(alg) == 1 }
 }
 
+pub unsafe fn PSA_ALG_IS_RSA_OAEP(alg: psa_algorithm_t) -> bool {
```
hug-dev, line 91: I think this one can be safe because it can only return `1` or `0` with no unspecified behaviour.

psa-crypto/src/types/key.rs
```diff
@@ -335,23 +365,46 @@ impl Attributes {
...
     fn export_key_output_size_base(key_type: Type, bits: usize) -> Result<usize> {
-        let size =
-            unsafe { psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(key_type.try_into()?, bits) };
-        if size > 0 {
-            Ok(size)
-        } else {
-            Err(Error::NotSupported)
+        match unsafe { psa_crypto_sys::PSA_EXPORT_KEY_OUTPUT_SIZE(key_type.try_into()?, bits) } {
+            0 => Err(Error::NotSupported),
+            size => Ok(size),
```
hug-dev, line 370: 🙆‍♂️

psa-crypto-sys/src/shim_methods.rs
sbailey-arm: I did wonder about this one. I put unsafe because although it can only return `0` or `1`, if you give it an unsupported input, the value it returns doesn't necessarily tell you that (because it can be either `0` or `1`), so you couldn't trust it.

hug-dev: I see what you mean that, you could give it a value that is not one of the recognised possible algorithm value for `psa_algorithm_t`, and the result would not really make sense.
In my opinion, `psa_algorithm_t` being just a typedef for `u32`, you are allowed to call this function with any `u32` value and for all these values, the function will always return `true` or `false` and never exhibit an `unsafe` behaviour.

I guess the problem is what you define as safe/unsafe :/

This is mostly bikeshedding at this point, so please ignore, it is not really important, and it is perfectly fine with `unsafe` 👌

PR #40, opened 2020-07-10
Added export_key

sbailey-arm:
Added export_key() along with a simple test.

### DIFF
--- a/psa-crypto/src/operations/key_management.rs
+++ b/psa-crypto/src/operations/key_management.rs
@@ -228,8 +228,60 @@ pub fn export_public(key: Id, data: &mut [u8]) -> Result<usize> {
         )
     })
     .to_result();
-    key.close_handle(handle)?;
+    let handle_close_res = key.close_handle(handle);
     export_res?;
+    handle_close_res?;
+    Ok(data_length)
+}
+
+/// Export a key pair in binary format
+///
+/// The key is written in `data`. The functions returns the number of bytes written.
+/// Please check the PSA Crypto API for a more complete description on the format of `data`.
+///
+/// # Example
+///
+/// ```
+/// # use psa_crypto::operations::key_management;
+/// # use psa_crypto::types::key::{Attributes, Type, Lifetime, Policy, UsageFlags};
+/// # use psa_crypto::types::algorithm::{AsymmetricSignature, Hash};
+/// # let mut attributes = Attributes {
+/// #     key_type: Type::RsaKeyPair,
+/// #     bits: 1024,
+/// #     lifetime: Lifetime::Volatile,
+/// #     policy: Policy {
+/// #         usage_flags: UsageFlags {
+/// #             sign_hash: true,
+/// #             sign_message: true,
+/// #             verify_hash: true,
+/// #             verify_message: true,
+/// #             export: true,
+/// #             ..Default::default()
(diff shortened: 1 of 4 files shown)


egrimley-arm:
Are you quietly fixing a bug in `export_public_key()` at the same time?

sbailey-arm:
> 
> 
> Are you quietly fixing a bug in `export_public_key()` at the same time?

I've been rumbled! 

I'm not sure it is actually a bug. We had a discussion about it a while ago, I think the result was 'it doesn't really matter' but returning the error for closing the key handle before the error for the actual operation doesn't seem right to me. Key handles should be going away at some point anyway.

egrimley-arm:
Might be good to add something like the following to the log message, then:

Also, minor improvement to export_public_key(): Return operation error in preference to close error.

psa-crypto/src/operations/key_management.rs
```diff
@@ -228,8 +228,59 @@ pub fn export_public(key: Id, data: &mut [u8]) -> Result<usize> {
...
+/// #     },
+/// # };
+/// psa_crypto::init().unwrap();
+/// let buffer_size = attributes.export_key_output_size().unwrap();
+/// let mut data = vec![0; buffer_size];
+/// let my_key = key_management::generate(attributes, None).unwrap();
+/// let size = key_management::export_public(my_key, &mut data).unwrap();
+/// data.resize(size, 0);
+/// ```
+pub fn export_pair(key: Id, data: &mut [u8]) -> Result<usize> {
```
hug-dev, line 272: I think the name here should match the PSA Crypto one: `export_key`

sbailey-arm: Ah yes, I did that before reading the description, doh. Fixed.

sbailey-arm:
> 
> 
> Might be good to add something like the following to the log message, then:
> 
> Also, minor improvement to export_public_key(): Return operation error in preference to close error.

I've added that to the commit message.

psa-crypto/tests/mod.rs
```diff
@@ -78,9 +78,45 @@ fn import_integration_test() {
     }
 }
 
+#[test]
+fn export_key_pair_test() {
```
ionut-arm, line 82: Would it be possible to have a test where you put a previously generated key file in the right place and export that key? It would help with more accurate testing (instead of just making sure you get _something_ on the other side). Would also help with testing in Parsec, as we won't have to be that thorough when testing there.

ionut-arm: I guess another option is to import and export a "static" key (i.e. a byte vector).

sbailey-arm: You mean like has been done in Parsec? Yes I can add that test here. I wasn't sure what kind of/how many tests are added to `psa-crypto` as there are only `generate` and `import` tests there currently.

ionut-arm: Yes, I think it makes more sense to put the tests at the lowest level where they apply - then neither Parsec nor other users of the crate need to test for that

sbailey-arm: Ah ok, good point. I've used the PK from Parsec's test, imported it, exported it, and verified the original an exported values match.

PR #43, opened 2020-07-17
Re-run build if any file under psa-crypto-sys/vendor has changed

### LINKED ISSUE #35: Changes to the C library do not trigger a rebuild
If the C code in `psa-crypto-sys/vendor/` is changed then `cargo` does not rebuild it.

A work-around is to run `touch psa-crypto-sys/build.rs` when the C code might have changed.

I think it would be better if `build.rs` found every file (and perhaps every directory) under `vendor/` (there's a crate called `walkdir`) and emitted `cargo:rerun-if-changed=PATH` for each one. (Or you could try to more selective, but persuading `cmake` to tell you which files really matter would probably be too much work.)

joechrisellis:
Fix #35

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This project uses the following third party crates:
 * cc (MIT and Apache-2.0)
 * log (MIT and Apache-2.0)
 * serde (MIT and Apache-2.0)
+* walkdir (MIT or UNLICENSE)
 
 ## Contributing
 
--- a/psa-crypto-sys/Cargo.toml
+++ b/psa-crypto-sys/Cargo.toml
@@ -18,6 +18,7 @@ links = "mbedcrypto"
 bindgen = "0.54.0"
 cc = "1.0.54"
 cmake = "0.1.44"
+walkdir = "2"
 
 [features]
 default = ["operations"]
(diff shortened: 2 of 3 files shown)


psa-crypto-sys/Cargo.toml
```diff
@@ -18,6 +18,7 @@ links = "mbedcrypto"
 bindgen = "0.54.0"
 cc = "1.0.54"
 cmake = "0.1.44"
+walkdir = "2"
```
hug-dev, line 21: We usually put the dependencies we use in the `README.md` file as well, for recognition and to check that they have a compatible license!

joechrisellis: Ahh gotcha, will fix. :)

README.md
```diff
@@ -30,6 +30,7 @@ This project uses the following third party crates:
 * cc (MIT and Apache-2.0)
 * log (MIT and Apache-2.0)
 * serde (MIT and Apache-2.0)
+* walkdir (MIT or UNLICENSE)
```
ionut-arm, line 33: Eh... is UNLICENSE compatible with Apache-2.0? 

I'm not sure I understand very well if it leans the GPL way: https://unlicense.org/

ionut-arm: Sounds like it's meant to be compatible with (almost?) any license - they just renounce their rights over the code altogether, so should be fine, but maybe we need to check

hug-dev: Got the confirmation it was ok to use.

PR #48, opened 2020-07-28
Added all missing ops listed in Service API that Mbed Crypto support …

sbailey-arm:
…(plus some it doesn't) found [here](https://parallaxsecond.github.io/parsec-book/service_api_coverage.html#single-part-operations).

The goal of this PR is to add all operations listed in the Single-part Service API table that are currently supported by Mbed Crypto. Since there are more than a couple of additions, here is a breakdown:

`psa_key_derivation_output_key` - supported, added, *
`psa_copy_key` - supported, added
`psa_purge_key` - unsupported, added
`psa_hash_compute` - supported, added
`psa_hash_compare` - supported, added
`psa_mac_compute` - unspported, added, ***
`psa_mac_verify` - unsupported, added, ***
`psa_cipher_encrypt` - unsupported, not added
`psa_cipher_decrypt` - unsupported, not added
`psa_aead_encrypt` - supported, added, ***
`psa_aead_decrypt` - supported, added, ***
`psa_sign_message` - unsupported, not added
`psa_verify_message` - unsupported, not added
`psa_key_derivation_key_agreement` - supported, added, *, **
`psa_raw_key_agreement` - supported, added

supported/unsupported - whether Mbed Crypto has support for this operation
added/not added - whether the operation has been added to psa-crypto in this PR

\* - these don't actually appear to be single-part operations and probably shouldn't be in that Single-part operation API table. However, `psa_key_derivation_output_key` was almost complete before @hug-dev  and I realised that so I completed them anyway.
** - `psa_key_derivation_key_agreement` appears as a separate operation in the API table. I took a different approach with it. It looks like just another input step so I've integrated it into the `psa_key_derivation_output_key` operation.

To give a quick overview of [what goes on with key derivation](https://armmbed.github.io/mbed-crypto/html/overview/functionality.html?#key-derivation) in the PSA spec, you have to create an operation struct and then call multiple `input` functions on it in order to give the operation data (either byte buffer or a key) to use at each step
in the derivation process (which steps you need to provide data for and the format required is dependant on the derivation algorithm. After you've set it up, you call `psa_key_derivation_output_key` to perform the actual derivation.

The general design of this particular multi-part operation is instead of having the user call multiple 'sub' operations, they provide all of the data upfront. A benefit of this is you can check before the operation is called
whether the correct data for the requested algorithm has been provided (or rather, you can't call it without providing the correct data). It started with just allowing byte buffers or key IDs, but as I mentioned, I expanded
it to include data required for a key agreement too.

*** - There is a potential problem with a truncated MAC. Creating a truncated MAC with the C macro is unspecified if 'mac_length is too small or too large for the specified MAC algorithm.'. The requirements for mac_length are 'This must be at most the full length of the MAC and must be at least an implementation-specified minimum. The implementation-specified minimum must not be zero.'.

The first requirement (must be at most the full length of the algorithm) should be easy enough to look up, but I can't find any indicators as to the min value that Mbed Crypto supports. The tests indicate that this value is 1, but I can't see any documentation for it.

We also don't currently do any checking on the values a user could put into our `Mac::Truncated` enum struct, which is what is passed to the macro. We could declare a separate Truncated struct, make the members private (outside of psa-crypto) to force users to use `Truncated::new` where we could do a runtime check on length, and have the enum hold that struct, but we still don't have anything to check against

The same is true for Aead. This should perhaps be listed as an issue, but since I noticed it yesterday and the PR was ready, I included it here.


ionut-arm:
While this one is somewhat okay, I strongly recommend adding them in smaller groups in the other repos, large PRs are very difficult to review.

sbailey-arm:
> 
> 
> While this one is somewhat okay, I strongly recommend adding them in smaller groups in the other repos, large PRs are very difficult to review.

Ok. I was asked if it'd be possible to do them in one to make it easier to review, but perhaps this is a bit overboard...! I'll cut them up into smaller groups in the rest of the repos to make them more bitesize (infact @joechrisellis might be taking on some of the operations (e.g. the hash operations) too).

ionut-arm:
My problem with large PRs is summarised by:


But then again, everyone has their style of reviewing! If the change is really simple it could work, I guess

sbailey-arm:
> 
> 
> My problem with large PRs is summarised by:
> 
> 
> But then again, everyone has their style of reviewing! If the change is really simple it could work, I guess

Yep I see your point 😄. Maybe I should try reviewing some PRs 😅! Anyway, these changes shouldn't be too complex (if the multi-part ops are, let me know and I'll pull them out) and I'll split it up for the rest of the repos 🙂.

psa-crypto/src/operations/key_management.rs
```diff
@@ -307,3 +307,99 @@ fn complete_new_key_operation(
...
+/// #         permitted_algorithms: AsymmetricSignature::RsaPkcs1v15Sign {
+/// #             hash_alg: Hash::Sha256.into(),
+/// #         }.into(),
+/// #     },
+/// # };
+/// psa_crypto::init().unwrap();
+/// let my_key = key_management::generate(attributes, None).unwrap();
+/// let my_key_copy = key_management::copy_key(my_key, attributes, None);
+/// ```
+pub fn copy_key(key_id_to_copy: Id, attributes: Attributes, id: Option<u32>) -> Result<Id> {
```
hug-dev, line 342: When thinking about it, we removed the "key" from all function names: this should probably be `copy`. Similar for `export_key` above, it should maybe be `export`. And `purge` below.

psa-crypto/src/operations/key_management.rs
```diff
@@ -307,3 +307,99 @@ fn complete_new_key_operation(
...
+/// psa_crypto::init().unwrap();
+/// let my_key = key_management::generate(attributes, None).unwrap();
+/// let size = key_management::purge_key(my_key).unwrap();
+/// ```*/
+/*
+pub fn purge_key(key_id: Id) -> Result<()> {
+    initialized()?;
+    let handle = key_id.handle()?;
+    Status::from(psa_crypto_sys::psa_purge_key(handle)).to_result()
+}*/
```
hug-dev, line 405: Maybe it is better to still keep the function but have an empty implementation for now? Using a log to notice it, with the same comment that you have above.

psa-crypto/src/operations/mac.rs
```diff
@@ -2,3 +2,134 @@
...
+/// # let mut attributes = Attributes {
+/// #     key_type: Type::RsaKeyPair,
+/// #     bits: 1024,
+/// #     lifetime: Lifetime::Volatile,
+/// #     policy: Policy {
+/// #         usage_flags: UsageFlags {
+/// #             sign_message: true,
+/// #             ..Default::default()
+/// #         },
+/// #         permitted_algorithms: Mac::FullLength(FullLengthMac::Hmac{hash_alg: Hash::Sha256}).into(),
```
hug-dev, line 34: Would be nice to have a `From<FullLength>` implementation on `FullLengthMac` to directly do:
```rust
FullLengthMac::Hmac{hash_alg: Hash::Sha256}.into()
```

psa-crypto/src/operations/mac.rs
```diff
@@ -2,3 +2,134 @@
...
+            input_message.as_ptr(),
+            input_message.len(),
+            expected_mac.as_ptr(),
+            expected_mac.len(),
+        )}
+    );
+    let close_handle_res = key_id.close_handle(key_handle);
+    let final_mac_verify_res = match mac_verify_res {
+        Status::Success => Ok(true),
+        Status::Error(Error::InvalidSignature) => Ok(false), // Indicates operation successful but compare did not match
```
hug-dev, line 130: Since this `Error` exists, maybe it is easier for consumers for this function to just return a `Result<()>`? If they want to check for actual signature failure, they can do a match for this specific error.

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+#[cfg(feature = "interface")]
+use super::status::{Error, Result, Status};
+#[cfg(feature = "interface")]
+use core::convert::{From, TryFrom};
+
+/// Key derivation operation for deriving keys from existing sources
+#[derive(Debug, Clone)]
+pub struct KeyDerivationOperation<'a> {
+    /// Key derivation algorithm and inputs
+    pub alg: KeyDerivationWithInputs<'a>,
```
hug-dev, line 17: `alg` could be named `inputs`, `KeyDerivationWithInputs` could be named `Inputs`

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+use super::algorithm::{Hash, KeyDerivation};
+use super::key::Id;
+#[cfg(feature = "interface")]
+use super::status::{Error, Result, Status};
+#[cfg(feature = "interface")]
+use core::convert::{From, TryFrom};
+
+/// Key derivation operation for deriving keys from existing sources
+#[derive(Debug, Clone)]
+pub struct KeyDerivationOperation<'a> {
```
hug-dev, line 15: Still about naming: would it be nicer if this file is named `key_derivation` and this structure `Operation`?

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+    /// Key derivation algorithm and inputs
+    pub alg: KeyDerivationWithInputs<'a>,
+    /// Maximum capacity of a key derivation operation
+    pub capacity: Option<usize>,
```
hug-dev, line 19: In the specs I can read:

hug-dev, line 19: In the specs I can read:

> For some algorithms, this step is mandatory because the output depends on the maximum capacity.

If we know for which ones this is mandatory, we can force it for them and leave it as `Option` for the other ones?

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+#[derive(Debug, Clone)]
+pub enum KeyDerivationWithInputs<'a> {
+    /// HKDF algorithm.
+    Hkdf {
+        /// A hash algorithm to use
+        hash_alg: Hash,
+        /// Salt, used in the "extract" step. It is optional; if omitted, the derivation uses an empty salt.
+        /// Typically a direct input, can also be a key of type `RawData`.
+        salt: Option<Input<'a>>,
+        /// Secret, used in the "extract" step. Must be `Input::Key(Id)` if used with `psa_key_derivation_output_key`.
```
hug-dev, line 32: You write in multiple places:

> Must be `Input::Key(Id)` if used with `psa_key_derivation_output_key`

Would there be a way to force that in the type system?

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+        /// Must not be larger than `PSA_TLS12_PSK_TO_MS_PSK_MAX_SIZE`.
+        secret: Input<'a>,
+        /// Label. Typically a direct input, can also be a key of type `RawData`.
+        label: Input<'a>,
+    },
+}
+
+/// Enumeration of the step of a key derivation.
+#[derive(Debug, Clone, Copy)]
+pub enum DerivationStep {
```
hug-dev, line 64: This could be `Step` 🧹

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+    pub capacity: Option<usize>,
+}
+
+/// Wrapper around KeyDerivation to enforce correct `Input`s
+#[derive(Debug, Clone)]
+pub enum KeyDerivationWithInputs<'a> {
+    /// HKDF algorithm.
+    Hkdf {
+        /// A hash algorithm to use
+        hash_alg: Hash,
```
hug-dev, line 28: Where is the hash alg actually used?

psa-crypto/src/operations/mac.rs
hug-dev, line 130: Since there is an `InvalidSignature` error, it could make sense to just return `Result<()>` from this function!

psa-crypto/src/types/key_derivation_operation.rs
hug-dev, line 17: alg could be named inputs, KeyDerivationWithInputs could be named Inputs

psa-crypto/src/types/key_derivation_operation.rs
hug-dev, line 15: Still about naming: would it be nicer if this file is named key_derivation and this structure Operation?

psa-crypto-sys/src/utils.rs
```diff
@@ -0,0 +1,15 @@
...
+use crate::types::psa_algorithm_t;
+/// Retrieves the tag length from an aead_alg.
+/// Note: `aead_alg` is an AEAD algorithm, such that `PSA_ALG_IS_AEAD(aead_alg)` is `true`.
+pub fn PSA_ALG_AEAD_TAG_TRUNCATED_LENGTH(aead_alg: psa_algorithm_t) -> usize {
+    const TAG_LENGTH_MASK: u8 = 0b00111111; // tag lengths are 6 bits in length
+    const PSA_V1_0_0_TAG_LENGTH_START_BIT: u32 = 16; // tag length at bit position [21:16]
+
+    let pre_mask_tag_length = aead_alg >> PSA_V1_0_0_TAG_LENGTH_START_BIT;
+
+    (pre_mask_tag_length as u8 & TAG_LENGTH_MASK) as usize
```
ionut-arm, line 14: Why is there a need for converting to `u8` since you convert to `usize` immediately after? I think just `& TAG_LENGTH_MASK` is enough to keep only the lowest 6 bits set.

ionut-arm:
Actually, a more primordial question - no tests?

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+            Status::from(unsafe {
+                psa_crypto_sys::psa_key_derivation_setup(
+                    &mut op,
+                    operation.alg.key_derivation().into(),
+                )
+            })
+            .to_result()?;
+            operation.alg.apply_inputs_to_op(&mut op)
+        } {
+            operation.abort(op)?;
```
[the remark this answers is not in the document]
sbailey-arm: The [key derivation setup](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?#c.psa_key_derivation_setup), [input bytes](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?highlight=psa_key_derivation_setup#c.psa_key_derivation_input_bytes) and [input key](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?highlight=psa_key_derivation_setup#c.psa_key_derivation_input_key) state that if they return an error, you have to call `abort`.

psa-crypto/src/types/algorithm.rs
```diff
@@ -497,6 +498,15 @@ pub enum KeyDerivation {
...
 
+/// Enumeration of supported input data for different input steps
+#[derive(Debug, Clone)]
+pub enum Input<'a> {
+    /// Byte input for key derivation
+    Bytes(&'a [u8]),
+    /// Key input for key derivation
+    Key(Id),
+}
+
```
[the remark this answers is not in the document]
sbailey-arm: It already is 🤔. Not sure how it ended up in both! I've removed the one in `algorithm.rs`

psa-crypto/src/types/algorithm.rs
```diff
@@ -755,6 +769,199 @@ impl From<AsymmetricEncryption> for psa_crypto_sys::psa_algorithm_t {
     }
 }
 
+#[cfg(feature = "interface")]
+impl TryFrom<psa_crypto_sys::psa_algorithm_t> for Mac {
+    type Error = Error;
+    fn try_from(alg: psa_crypto_sys::psa_algorithm_t) -> Result<Self> {
+        if psa_crypto_sys::PSA_ALG_IS_FULL_LENGTH_MAC(alg) {
+            Ok(Mac::FullLength(alg.try_into()?))
+        } else {
```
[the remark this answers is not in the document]
sbailey-arm: Bit confused what you mean here. If it ends up in this method to begin with, it _has_ to be a MAC algorithm, otherwise the code above this method isn't working correctly. The block starting on line 544 (the `TryFrom<psa_crypto_sys::psa_algorithm_t> for Algorithm` is what directs MAC algorithms to this particular method.

psa-crypto/src/types/algorithm.rs
```diff
@@ -755,6 +769,199 @@ impl From<AsymmetricEncryption> for psa_crypto_sys::psa_algorithm_t {
...
+
+#[cfg(feature = "interface")]
+impl From<Mac> for psa_crypto_sys::psa_algorithm_t {
+    fn from(mac: Mac) -> Self {
+        match mac {
+            Mac::FullLength(full_length_mac) => full_length_mac.into(),
+            Mac::Truncated {
+                mac_alg: alg,
+                mac_length: length,
+                // The following call is NOT currently checked. If length is invalid, the return of this call is unspecified
```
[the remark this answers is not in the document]
sbailey-arm: Agreed 🙂.

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+
+//! # PSA Operation types
```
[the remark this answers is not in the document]
sbailey-arm: Changed to `PSA Key Derivation Operation Types`.

psa-crypto/src/operations/key_derivation.rs
```diff
@@ -2,3 +2,95 @@
...
+    let mut key_attributes = psa_crypto_sys::psa_key_attributes_t::try_from(attributes)?;
+    let id = if let Some(id) = id {
+        unsafe { psa_crypto_sys::psa_set_key_id(&mut key_attributes, id) };
+        id
+    } else {
+        0
+    };
+    let mut handle_for_new_key = 0;
+
+    let mut op: psa_crypto_sys::psa_key_derivation_operation_t = operation.clone().try_into()?;
```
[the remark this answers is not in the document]
sbailey-arm: The reason for not implementing `Copy` on this was to stop the user ending up with an `operation` that they could use after it was aborted. The current way it is setup, once the user calls the key derivation operation, they cannot use that `operation` again.

When you call abort of the C struct, it resets it to default values, which are not 'valid'. I didn't want to make the types on `operation` `Option<T>` as I don't think it fits, so there is no way to reset the Rust `operation` struct, you have to start again, specifying the inputs you want at each stage for whatever algorithm you're using.

Another way of doing it would be to allow the user to keep their `operation` struct and re-use it. Its just with the `abort` in Mbed Crypto, I was trying to replicate it. Now that I think about it, I can't think of a reason why it shouldn't be done like that.

This is more of a Rust thing but I find that `clone` a bit annoying. I played around with making the `TryFrom` method take a reference, but then it got upset when I had the `abort` there as there was still a living borrow. I get why its complaining, but I know the borrow isn't used (e.g. stored) for use after `TryFrom` returns!

psa-crypto/src/operations/key_derivation.rs
```diff
@@ -2,3 +2,95 @@
...
+    Status::from(unsafe {
+        psa_crypto_sys::psa_key_derivation_output_key(
+            &key_attributes,
+            &mut op,
+            &mut handle_for_new_key,
+        )
+    })
+    .to_result()?;
+    operation.abort(op)?;
```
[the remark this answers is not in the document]
sbailey-arm: Yes. If you look at the last [key derivation step](https://armmbed.github.io/mbed-crypto/html/overview/functionality.html#key-derivation), you have to call `abort` to clean up its resources.

psa-crypto/src/operations/key_management.rs
sbailey-arm: I've commented out the code and the example so now all it does is use the `error` macro. Is this what you meant?

psa-crypto/src/operations/key_derivation.rs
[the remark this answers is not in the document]
ionut-arm: > to stop the user ending up with an operation that they could use after it was aborted.

They can still clone it, right? An alternative is to implement your own internal clone that's only visible to this crate.

psa-crypto/src/types/key_derivation_operation.rs
sbailey-arm: I did look into doing exactly that but couldn't find any documentation that states for which algorithms `capacity` is mandatory 😞.

psa-crypto/src/types/key_derivation_operation.rs
sbailey-arm: I couldn't think of a clean way of doing it. An option would be to have it as a parameter on the method to force `Input::Key(Id)` but that seems a bit janky. I can have another think about it after everything else is sorted.

psa-crypto/src/types/key_derivation_operation.rs
ionut-arm: I don't know about this one - not sure if `Step` is so common a term in this kind of context to be fully clear what it means without the `Derivation` part of the name

psa-crypto/src/types/key_derivation_operation.rs
sbailey-arm: When converting to `FullLengthMac::Hkdf` as it takes a hash algorithm (as does the [C macro](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html?#c.PSA_ALG_HKDF) to build a Hkdf derivation algorithm).

psa-crypto/src/operations/mac.rs
sbailey-arm: I originally did that but thought since the operation is 'successful' in the event of an `InvalidSignature` error, it should tell the user that the operation completed successfully but the comparison did not match. I guess it doesn't actually matter that much and just returning the error is slightly simpler 🤔.

psa-crypto/src/types/key_derivation_operation.rs
sbailey-arm: I think so, the shorter file name still conveys what's there. Just calling it `Operation` _could_ be confusing for people if they have multiple different types of `Operation` in the same place, but they can just rename it themselves in the import statement.

psa-crypto/src/operations/mac.rs
ionut-arm: I guess `Result<Result<()>>` is more correct, but fugly

psa-crypto-sys/src/utils.rs
sbailey-arm: Originally had that but compiler said no - 
```
14 |     (pre_mask_tag_length & TAG_LENGTH_MASK) as usize
   |                            ^^^^^^^^^^^^^^^ expected `u32`, found `u8`

```

ionut-arm: And if you make `const TAG_LENGTH_MASK: u32 = 0b111111` ?

psa-crypto-sys/src/c/shim.c
```diff
@@ -102,6 +108,21 @@ shim_PSA_ALG_IS_MAC(psa_algorithm_t alg) {
...
+}
+
+int
+shim_PSA_ALG_IS_BLOCK_CIPHER_MAC (psa_algorithm_t alg) {
+    return PSA_ALG_IS_BLOCK_CIPHER_MAC (alg);
+}
+
+int
+shim_PSA_ALG_IS_FULL_LENGTH_MAC (psa_algorithm_t alg) {
+    return PSA_ALG_IS_BLOCK_CIPHER_MAC (alg);
```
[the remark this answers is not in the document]
sbailey-arm: Yes, I think that is an error. The conversion code is going to need re-jigging if we don't have access to a macro that does something like the name of this one suggests 🤔.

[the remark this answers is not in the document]
ionut-arm: There is a macro `PSA_ALG_FULL_LENGTH_MAC` that converts any MAC algorithm (truncated or not) to its full length version - you could return something like `PSA_ALG_FULL_LENGTH_MAC(alg) == alg`?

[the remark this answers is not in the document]
sbailey-arm: Ah yes, that's nice an elegant!

psa-crypto/src/operations/key_derivation.rs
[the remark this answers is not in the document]
sbailey-arm: > Or a TryFrom<&KeyDerivationOperation> if that's possible, I think that's a thing

That's what I was referring to in this comment. You can have a `TryFrom` with a borrow, but that doesn't solve this problem ☹️.

> This is more of a Rust thing but I find that clone a bit annoying. I played around with making the TryFrom method take a reference, but then it got upset when I had the abort there as there was still a living borrow. I get why its complaining, but I know the borrow isn't used (e.g. stored) for use after TryFrom returns!

psa-crypto/src/types/algorithm.rs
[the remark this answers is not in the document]
ionut-arm: > it _has_ to be a MAC algorithm

`impl`s are "public", though - if someone has a `psa_algorithm_t` and they do `Mac::try_from(alg)`, this gets called, regardless of what that `alg` was.

psa-crypto/src/types/key_derivation_operation.rs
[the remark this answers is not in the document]
hug-dev: I see. That would mean that we have to add `abort` calls above in `psa_key_derivation_setup` fails

psa-crypto/src/operations/hash.rs
```diff
@@ -0,0 +1,85 @@
...
+            hash_alg.into(),
+            input.as_ptr(),
+            input.len(),
+            hash_to_compare.as_ptr(),
+            hash_to_compare.len(),
+        )
+    }) {
+        Status::Success => Ok(true),
+        Status::Error(Error::InvalidSignature) => Ok(false), // Indicates operation successful but compare did not match
+        Status::Error(error) => Err(error),
```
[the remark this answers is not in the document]
sbailey-arm: Changed it to return `Result<()>` as with the other occurrence above.

psa-crypto/src/operations/hash.rs
```diff
@@ -0,0 +1,85 @@
...
+///
+/// # Example
+///
+/// ```
+/// # use psa_crypto::operations::hash::{hash_compute, hash_compare};
+/// # use psa_crypto::types::algorithm::Hash;
+/// # const MESSAGE: [u8; 32] = [
+/// #     0x69, 0x3E, 0xDB, 0x1B, 0x22, 0x79, 0x03, 0xF4, 0xC0, 0xBF, 0xD6, 0x91, 0x76, 0x37, 0x84, 0xA2,
+/// #     0x94, 0x8E, 0x92, 0x50, 0x35, 0xC2, 0x8C, 0x5C, 0x3C, 0xCA, 0xFE, 0x18, 0xE8, 0x81, 0x37, 0x78,
+/// # ];
```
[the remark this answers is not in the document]
sbailey-arm: Not really! I'm not sure which bits should be hidden and which shouldn't. I was trying to make all 'obvious' code hidden with the less obvious code not. However, what 'obvious' actually means will depend on who you ask. E.g. I tried to make sure the bits that took a while to get working (e.g. correct key types, sizes, policy flags and algorithms) weren't hidden.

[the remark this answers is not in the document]
ionut-arm: I think it makes sense to hide the actual values/buffers, as the type really isn't something unusual or hard to grasp. Maybe the imports are actually useful is someone wants to just copy-paste code straight from the example, but yeah, really not sure.

As long as they're consistent, choose whatever makes sense to you!

psa-crypto/src/operations/key_derivation.rs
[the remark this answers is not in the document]
hug-dev: What will  happen for the slices when cloning? Will that make deep copies of them?

psa-crypto/src/types/key_derivation_operation.rs
```diff
@@ -0,0 +1,265 @@
...
+
+impl KeyDerivationOperation<'_> {
+    /// Clears operation C struct and consumes KeyDerivationOperation struct
+    #[cfg(feature = "interface")]
+    pub(crate) fn abort(
+        self,
+        mut op: psa_crypto_sys::psa_key_derivation_operation_t,
+    ) -> Result<()> {
+        Status::from(unsafe { psa_crypto_sys::psa_key_derivation_abort(&mut op) }).to_result()
+    }
```
ionut-arm, line 264: Why is this a method (instead of just a function on the type)? `self` isn't used anywhere, it would remove your need to `clone` the instance in some places

psa-crypto/src/types/key_derivation_operation.rs
ionut-arm: I'm not sure I follow... `KeyDerivationOperation` is only usable with `psa_key_derivation_output_key` - or is there more to come?

psa-crypto/src/types/algorithm.rs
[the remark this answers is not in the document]
sbailey-arm: Ah, I see what you mean. There should be a fail case for `TryFrom`, otherwise you'd use `From`.

psa-crypto/src/types/key_derivation_operation.rs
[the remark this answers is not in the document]
sbailey-arm: It should do already, I think? The call to `psa_key_derivation_setup` and to `apply_inputs_to_op` (that method sort out calling the correct `input` functions) are wrapped in the same block that listens for an error. I was going to do it separately, then realised I could block them together.

psa-crypto/src/operations/key_derivation.rs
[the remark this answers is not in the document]
sbailey-arm: Hmmm good question. Slice [doesn't seem to implement](https://doc.rust-lang.org/std/primitive.slice.html#trait-implementations) `Clone`. It could copy the reference but then you could end up with a dangling pointer, which Rust doesn't allow 🤔.

psa-crypto/src/types/key_derivation_operation.rs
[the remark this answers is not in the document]
hug-dev: What about the error line 238, can't we escape the function without `abort` if we leave it at:
```rust
.to_result()?
```
?

psa-crypto/src/types/key_derivation_operation.rs
sbailey-arm: So it consumes `self`, stopping the user from trying to reuse something that has been cleared. Think it might be better just to do it like `attributes`, let the user keep their un-cleared Rust `operation` struct while clearing the C version. I think it deviates from the PSA spec as the struct is supposed to be cleared (and we can't do that for reasons already mentioned) but I can't think of any major problem with it.

sbailey-arm:
I've pulled `key_derivation` out of the PR and will make it a separate one.

sbailey-arm:
> 
> 
> Actually, a more primordial question - no tests?

Not yet. Didn't want to make this PR even bigger, but can add them here if required.

ionut-arm:
> 
> Not yet. Didn't want to make this PR even bigger, but can add them here if required.

No, another PR with tests makes sense

psa-crypto-sys/src/utils.rs
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Contributors to the Parsec project.
+// SPDX-License-Identifier: Apache-2.0
+#![allow(non_snake_case)]
+/// Additional functionality required that Mbed Crypto does not provide
```
hug-dev, line 4: Maybe it is "PSA Crypto" does not provide?
Also it might be good to rename this file `extra.rs`. `utils.rs` often ends up being a pot-pourri of non-sorted things 😃

psa-crypto/src/operations/hash.rs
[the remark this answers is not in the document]
hug-dev: Did you 👀 ? Sorry if you did not push everything yet 😅

psa-crypto/src/operations/mac.rs
```diff
@@ -2,3 +2,129 @@
...
+/// let mut mac = vec![0; buffer_size];
+///
+/// let size = compute_mac(my_key,
+///                     mac_alg,
+///                      &MESSAGE,
+///                      &mut mac).unwrap();
+/// mac.resize(size, 0);
+/// assert!(verify_mac(my_key, mac_alg, &MESSAGE, &mac));
+/// ```
+pub fn verify_mac(key_id: Id, mac_alg: Mac, input_message: &[u8], expected_mac: &[u8]) -> Result<bool> {
```
hug-dev, line 112: I guess this also should return `Result<()>`

psa-crypto/src/operations/mac.rs
```diff
@@ -2,3 +2,129 @@
...
+            mac_alg.into(),
+            input_message.as_ptr(),
+            input_message.len(),
+            expected_mac.as_ptr(),
+            expected_mac.len(),
+        )}
+    ).to_result();
+    let close_handle_res = key_id.close_handle(key_handle);
+    mac_verify_res?;
+    close_handle_res
```
hug-dev, line 129: How can this even compile ❓ Isn't `close_handle_res` of type `Result<()>`?

hug-dev: Ah, it's not added as a module because Mbed Crypto does not support it!

psa-crypto/src/operations/mac.rs
sbailey-arm: Yeah, sorry, I have changed these but my virtual machine has been taking the last few days off, must have been in its saved state which was lost. I'll re-change them 🙂

psa-crypto-sys/src/utils.rs
sbailey-arm: That works

PR #49, opened 2020-08-03
Add key derivation

sbailey-arm:
This PR is based on #48 as it uses some conversion code added in there.

Relevant notes from that PR: `psa_key_derivation_key_agreement` appears as a separate operation in the API table. I took a different approach with it. It looks like just another input step so I've integrated it into the psa_key_derivation_output_key operation.

To give a quick overview of what goes on with key derivation in the PSA spec, you have to create an operation struct and then call multiple input functions on it in order to give the operation data (either byte buffer or a key) to use at each step
in the derivation process (which steps you need to provide data for and the format required is dependant on the derivation algorithm. After you've set it up, you call psa_key_derivation_output_key to perform the actual derivation.

The general design of this particular multi-part operation is instead of having the user call multiple 'sub' operations, they provide all of the data upfront. A benefit of this is you can check before the operation is called
whether the correct data for the requested algorithm has been provided (or rather, you can't call it without providing the correct data). It started with just allowing byte buffers or key IDs, but as I mentioned, I expanded
it to include data required for a key agreement too.


sbailey-arm:
@hug-dev brought [this](https://github.com/parallaxsecond/rust-psa-crypto/pull/48#discussion_r464283863) up in the other PR (that used to contain this addition).

That is correct, I've tested it and it doesn't reach the `abort` code 😞 (I thought it would throw the error to the next scope up, not the next function up). I've updated the code to something that does work.

sbailey-arm:
I've made some changes to this now. Aborting the operation works like attributes - you pass `Operation::abort` the C struct and you get to keep your Rust one (can be reused).

I've also added conversions from `psa_algorithm_t` to `KeyAgreement` and `KeyDerivation`. They aren't needed (which is why I've only just done it!) but added for completion.

sbailey-arm:
(it'll be easier to review this PR once #48 has been merged to get rid of a load of the differences - this PR depends on some stuff in #48 which is why I haven't completely separated them)

ionut-arm:
I think you'll need to rebase on top of latest `master`, since the changes from the `all-ops` commit still appear

sbailey-arm:
> 
> 
> I think you'll need to rebase on top of latest `master`, since the changes from the `all-ops` commit still appear

Done 🙂. Now only the actual changes appear.

psa-crypto/src/operations/key_derivation.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    let mut op: psa_crypto_sys::psa_key_derivation_operation_t = operation.try_into()?;
+    Status::from(unsafe {
+        psa_crypto_sys::psa_key_derivation_output_key(
+            &key_attributes,
+            &mut op,
+            &mut handle_for_new_key,
+        )
+    })
+    .to_result()?;
```
hug-dev, line 89: I think we discussed that on the previous PR, but is aborting `op` necessary here if line 89 fails?

psa-crypto/src/types/algorithm.rs
```diff
@@ -947,6 +987,13 @@ impl From<RawKeyAgreement> for psa_crypto_sys::psa_algorithm_t {
     }
 }
 
+#[cfg(feature = "interface")]
```
hug-dev, line 990: Does it need this feature?

psa-crypto/src/types/key_derivation.rs
```diff
@@ -0,0 +1,257 @@
...
+pub enum Inputs<'a> {
+    /// HKDF algorithm.
+    Hkdf {
+        /// A hash algorithm to use
+        hash_alg: Hash,
+        /// Salt, used in the "extract" step. It is optional; if omitted, the derivation uses an empty salt.
+        /// Typically a direct input, can also be a key of type `RawData`.
+        salt: Option<Input<'a>>,
+        /// Secret, used in the "extract" step. This is typically a key of type `Derive` , or the shared secret
+        /// resulting from a key agreement obtained via psa_key_derivation_key_agreement().
```
hug-dev, line 33: Can you please replace the PSA names with the corresponding items in this repo?

psa-crypto/src/operations/key_derivation.rs
```diff
@@ -0,0 +1,92 @@
...
+/// #             derive: true,
+/// #             ..Default::default()
+/// #         },
+/// #         permitted_algorithms: KeyDerivation::Hkdf {
+/// #            hash_alg: Hash::Sha256,
+/// #         }.into()
+/// #     }
+/// # };
+///
+/// # let mut derived_key_attributes = Attributes {
```
hug-dev, line 40: Can the `key_type` and the `permitted_algorithms` of the derived key can be anything? Could it be used for RSA signing for example?

psa-crypto/src/operations/key_derivation.rs
```diff
@@ -0,0 +1,92 @@
...
+/// The key's location, usage policy, type and size are taken from attributes.
+/// # Example
+///
+/// ```
+/// use psa_crypto::operations::{key_derivation, key_management};
+/// use psa_crypto::types::key::{Attributes, Type, Lifetime, Policy, UsageFlags};
+/// use psa_crypto::types::algorithm::{Hash, KeyDerivation};
+/// use psa_crypto::types::key_derivation::{Operation, Inputs, Input};
+///
+/// # const KEY_DATA: [u8; 23] = [0; 23];
```
hug-dev, line 24: In your example, could this data come from the key agreement algorithm?

psa-crypto/src/operations/key_derivation.rs
sbailey-arm: Yes, in all cases (including success) other than and `InsifficientData` error being returned.

psa-crypto/src/types/algorithm.rs
sbailey-arm: Nope, good spot. This has already been added in #51 without it so I can remove it here and rebase onto master.

psa-crypto/src/operations/key_derivation.rs
sbailey-arm: I think you'd use the output of a key agreement algorithm as an input to the key derivation step `secret` as an alternative to creating a new key to be used in the `secret` input step. Whether they result in the same outcome, I'm not sure.

hug-dev: I think it would be nice for the example to show the key agreement + key derivation example!

psa-crypto/src/operations/key_derivation.rs
sbailey-arm: I believe so. Looking at the [description for `psa_key_derivation_output_key`](https://armmbed.github.io/mbed-crypto/html/api/ops/kdf.html#c.psa_key_derivation_output_key), there doesn't seem to be a restriction on the type of key. At the end it says 'For other key types', so presumably allows all types.

PR #51, opened 2020-08-06
Added conversion helper methods

sbailey-arm:


psa-crypto/src/types/algorithm.rs
```diff
@@ -925,6 +924,12 @@ impl From<AeadWithDefaultLengthTag> for psa_crypto_sys::psa_algorithm_t {
     }
 }
 
+impl From<AeadWithDefaultLengthTag> for Aead {
```
hug-dev, line 927: 🙆‍♂️

PR #52, opened 2020-08-07
Refactor of key_derivation

sbailey-arm:
Currently, key derivation does not properly support key agreement. This PR is a possible solution.

### DIFF
--- a/psa-crypto/src/types/key_derivation.rs
+++ b/psa-crypto/src/types/key_derivation.rs
@@ -3,7 +3,7 @@
 
 //! # PSA Key Derivation Operation types
 
-use super::algorithm::{Hash, KeyDerivation};
+use super::algorithm::{Hash, KeyDerivation, RawKeyAgreement};
 use super::key::Id;
 #[cfg(feature = "interface")]
 use super::status::{Error, Result, Status};
@@ -32,7 +32,7 @@ pub enum Inputs<'a> {
         /// Secret, used in the "extract" step. This is typically a key of type `Derive` , or the shared secret
         /// resulting from a key agreement, using `Input::KeyAgreement`.
         /// Must be a key or key agreement input if used with `psa_key_derivation_output_key`.
-        secret: Input<'a>,
+        secret: InputSecret<'a>,
         /// Info, used in the "expand" step. Typically a direct input, can also be a key of type `RawData`.
         info: Input<'a>,
     },
@@ -45,7 +45,7 @@ pub enum Inputs<'a> {
         /// Secret, used in the "extract" step. This is typically a key of type `Derive` , or the shared secret
         /// resulting from a key agreement, using `Input::KeyAgreement`.
         /// Must be a key or key agreement input if used with `psa_key_derivation_output_key`.
-        secret: Input<'a>,
+        secret: InputSecret<'a>,
         /// Label. Typically a direct input, can also be a key of type `RawData`.
         label: Input<'a>,
     },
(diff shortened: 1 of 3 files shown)


sbailey-arm:
Example isn't complete so this won't pass CI yet.

psa-crypto/src/types/key_derivation.rs
```diff
@@ -32,7 +32,7 @@ pub enum Inputs<'a> {
         /// Secret, used in the "extract" step. This is typically a key of type `Derive` , or the shared secret
         /// resulting from a key agreement, using `Input::KeyAgreement`.
         /// Must be a key or key agreement input if used with `psa_key_derivation_output_key`.
```
sbailey-arm, line 34: Is it worth duplicating these structs to enforce this? So we'd have an operation struct for `output_key` and one for `output_bytes` where the `output_key` struct only allows a key or key agreement for the `secret` step?

ionut-arm: No, I think that's overkill, as I assume most of the code would have to be duplicated too, right?

sbailey-arm: Yes most of it would have to be duplicated.

ionut-arm: I'd say writing a few lines to check the validity is better than duplicating tens/hundreds to make ensure that statically. If we have a regression test for it that should be enough. Do you think the static checking would bring a big benefit?

sbailey-arm: Yes, you're probably right. If someone wants to use this on an embedded device, they may end up appreciating the code size savings over enforcing correct parameters.

> Do you think the static checking would bring a big benefit?
Not big, no. It's easy enough to just read the documentation! But if there's a cheap enough way of forcing something to be correct, I'd take it (haven't though of a way to enforce this ).

sbailey-arm:
mbedTLS doesn't support key output if key agreement is used as the input for the `secret` step. The commented out examples shows how it could be used, once this has been added ([here](https://github.com/ARMmbed/mbedtls/issues/3550) is the issue).

sbailey-arm:
Added two tests. One that uses a key for input of secret which runs and passes and a second that uses key agreement as input for secret which is ignored as it will not currently pass.

ionut-arm:
You seem to have added some commits with the original derivation/agreement work

sbailey-arm:
>You seem to have added some commits with the original derivation/agreement work

I've rebased this onto the branch that I added all of the tests to. Plan was to have the other PR merged to main, then this should show the changes to key derivation.

psa-crypto/src/types/key_derivation.rs
```diff
@@ -207,11 +222,20 @@ pub enum Input<'a> {
     Bytes(&'a [u8]),
     /// Key input for key derivation
     Key(Id),
+}
+
+/// Enumeration of supported input data for different input steps
+#[derive(Debug, Clone, Copy)]
+pub enum InputSecret<'a> {
```
ionut-arm, line 229: I quite like the idea of separating the `KeyAgreement` out! Could you also implement a `TryFrom<Input> for InputSecret`, though, just so it's easier to just use the `Input`?

psa-crypto/src/types/algorithm.rs
```diff
@@ -992,12 +992,6 @@ impl From<RawKeyAgreement> for psa_crypto_sys::psa_algorithm_t {
     }
 }
 
-impl From<RawKeyAgreement> for KeyAgreement {
-    fn from(raw_key_agreement: RawKeyAgreement) -> Self {
-        KeyAgreement::Raw(raw_key_agreement)
-    }
-}
```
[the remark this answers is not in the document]
sbailey-arm: Don't think so but I've re-added it, makes sense to keep it imo.

psa-crypto/src/types/key_derivation.rs
sbailey-arm: Yep, can do a `From` (full of lovely lifetime specifiers...!)

ionut-arm:
Thanks! Apologies for taking so long to get this in

PR #54, opened 2020-08-11
Added helper methods for checking derivation policy flag

sbailey-arm:


sbailey-arm:
> 
> 
> Actually - unit tests? I'm always confused about whether we want tests for this kinda thing - code is trivial but maybe will be changed in the future?

Added a test for `is_derivable`, is that what you meant? Seem there are only tests for 3 other usage flags.

PR #55, opened 2020-08-17
Added helper methods and changed safety of size macros

sbailey-arm:
Added temporary implementation for getting output size of shared secret with EDCH algorithm (will be replaced with call to mbed TLS once it supports the macro).

Changed safety of some output size macro shim declarations as they have undefined return.

### DIFF
--- a/psa-crypto-sys/src/extras.rs
+++ b/psa-crypto-sys/src/extras.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 #![allow(non_snake_case)]
 /// Additional functionality required that PSA Crypto does not provide
-use crate::types::psa_algorithm_t;
+use crate::types::{psa_algorithm_t, psa_key_type_t};
 
 /// Retrieves the tag length from an aead_alg.
 /// Note: `aead_alg` is an AEAD algorithm, such that `PSA_ALG_IS_AEAD(aead_alg)` is `true`.
@@ -13,6 +13,27 @@ pub fn PSA_ALG_AEAD_TAG_TRUNCATED_LENGTH(aead_alg: psa_algorithm_t) -> usize {
     let pre_mask_tag_length = aead_alg >> PSA_V1_0_0_TAG_LENGTH_START_BIT;
 
     (pre_mask_tag_length & TAG_LENGTH_MASK) as usize
+}
+
+/// Retrieves the output size of an ECDH raw key agreement operation shared secret.
+/// Caller must ensure key type is compatible.
+/// Returns 0 if key size is too large.
+/// This does not match any PSA macro, it will be replaces by PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE once
+/// mbedTLS adds support for it.
+pub unsafe fn PSA_RAW_ECDH_KEY_AGREEMENT_OUTPUT_SIZE(
+    _key_type: psa_key_type_t,
+    key_bits: usize,
+) -> usize {
+    /*
+    The size of the shared secret is always `ceiling(m/8)` bytes long where `m` is the bit size associated with the curve,
+    i.e. the bit size of the order of the curve's coordinate field. When m is not a multiple of 8, the byte containing the most
+    significant bit of the shared secret is padded with zero bits.
+    */
+    if let Some(numerator) = key_bits.checked_add(7) {
+        numerator / 8
+    } else {
+        0
+    }
 }
 
 #[test]
--- a/psa-crypto/src/types/key.rs
+++ b/psa-crypto/src/types/key.rs
@@ -6,9 +6,9 @@
 #![allow(deprecated)]
 #[cfg(feature = "operations")]
 use crate::initialized;
-use crate::types::algorithm::{Algorithm, Cipher, KeyAgreement, RawKeyAgreement};
 #[cfg(feature = "interface")]
-use crate::types::algorithm::{AsymmetricEncryption, AsymmetricSignature, Mac};
+use crate::types::algorithm::{Aead, AsymmetricEncryption, AsymmetricSignature, Mac};
+use crate::types::algorithm::{Algorithm, Cipher, KeyAgreement, RawKeyAgreement};
 #[cfg(feature = "operations")]
 use crate::types::status::Status;
 use crate::types::status::{Error, Result};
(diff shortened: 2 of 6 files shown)


psa-crypto-sys/src/extras.rs
```diff
@@ -15,6 +15,22 @@ pub fn PSA_ALG_AEAD_TAG_TRUNCATED_LENGTH(aead_alg: psa_algorithm_t) -> usize {
...
+pub unsafe fn PSA_RAW_ECDH_KEY_AGREEMENT_OUTPUT_SIZE(
+    _key_type: psa_key_type_t,
+    key_bits: usize,
+) -> usize {
+    /*
+    The size of the shared secret is always `ceiling(m/8)` bytes long where `m` is the bit size associated with the curve,
+    i.e. the bit size of the order of the curve's coordinate field. When m is not a multiple of 8, the byte containing the most
+    significant bit of the shared secret is padded with zero bits.
+    */
+    (key_bits - 1) / 9 // Round the division up
```
hug-dev, line 31: What if `key_bits` is `0`? This might panic in debug mode and overflow in release mode. I propose the following:
```
(key_bits + 7) / 8
```
or even better using [`checked_add`](https://doc.rust-lang.org/std/primitive.usize.html#method.checked_add) for overflows. We can choose and document to return `0` if an overflow would happen.

psa-crypto-sys/src/extras.rs
```diff
@@ -15,6 +15,22 @@ pub fn PSA_ALG_AEAD_TAG_TRUNCATED_LENGTH(aead_alg: psa_algorithm_t) -> usize {
     (pre_mask_tag_length & TAG_LENGTH_MASK) as usize
 }
 
+/// Retrieves the output size of an ECDH raw key agreement operation shared secret.
+/// Caller must ensure key type is compatible.
+/// This does not match any PSA macro, it will be replaces by PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE once
+/// mbedTLS adds support for it.
+pub unsafe fn PSA_RAW_ECDH_KEY_AGREEMENT_OUTPUT_SIZE(
+    _key_type: psa_key_type_t,
```
hug-dev, line 23: Are you keeping it because `PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE` will use it?

psa-crypto/src/types/key.rs
```diff
@@ -428,12 +428,41 @@ impl Attributes {
         })
     }
 
-    /// Sufficient buffer size for the MAC of the specified algorithm, if compatible
+    /// Sufficient buffer size for an encrypted message using the given aead algorithm
     #[cfg(feature = "interface")]
-    pub fn mac_length(self, mac_alg: Mac) -> Result<usize> {
-        self.compatible_with_alg(mac_alg.into())?;
```
hug-dev, line 434: Are you sure no-one is using this method?

psa-crypto/src/types/key.rs
```diff
@@ -428,12 +428,41 @@ impl Attributes {
...
+            )
+        })
+    }
+
+    /// Sufficient buffer size for the resulting shared secret from a raw key agreement
+    #[cfg(feature = "interface")]
+    pub fn raw_key_agreement_output_size(self, alg: RawKeyAgreement) -> Result<usize> {
+        self.compatible_with_alg(KeyAgreement::Raw(alg).into())?;
+        Ok(unsafe {
+            psa_crypto_sys::PSA_RAW_ECDH_KEY_AGREEMENT_OUTPUT_SIZE(
```
hug-dev, line 462: Should we check that `alg` Is only ECDH?

psa-crypto-sys/src/extras.rs
sbailey-arm: I was told by the psa crypto api guys to only bother implementing the ECDH one as Mbed TLS will gain the full macro before it gains support for FFDH. When Mbed TLS adds the macro, this will be removed from `psa-crypto-sys`.

psa-crypto-sys/src/extras.rs
sbailey-arm: Didn't think of that. Though, will a valid key ever have 0 bits? I like your proposal, nice and fast 🚀

psa-crypto/src/types/key.rs
sbailey-arm: Yes 👀

psa-crypto-sys/src/extras.rs
hug-dev: oh I meant specifically about the `_key_type` parameter that is not used, sorry.

sbailey-arm: aaah ok. Yes you are right, `PSA_RAW_KEY_AGREEMENT_OUTPUT_SIZE` does require `key_type` so I thought I'd keep the signature the same (even though the name is slightly different).

psa-crypto/src/types/key.rs
```diff
@@ -436,6 +436,47 @@ impl Attributes {
...
+        })
+    }
+
+    /// Sufficient buffer size for an encrypted message using the given aead algorithm
+    #[cfg(feature = "interface")]
+    pub fn aead_decrypt_output_size(self, alg: Aead, ciphertext_len: usize) -> Result<usize> {
+        self.compatible_with_alg(alg.into())?;
+        Ok(unsafe {
+            psa_crypto_sys::PSA_AEAD_DECRYPT_OUTPUT_SIZE(
+                /*self.key_type.try_into() PSA API specifies including this parameter*/
```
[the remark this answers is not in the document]
sbailey-arm: I thought it'd make it easier to see what's going on in the future when Mbed TLS is brought up to date with the spec. If it stops compiling, the person looking into it should see it and has a strong lead? I can remove it if its redundant or creates clutter.

psa-crypto/src/types/key.rs
```diff
@@ -436,6 +436,47 @@ impl Attributes {
...
+                alg.into(),
+                plaintext_len,
+            )
+        })
+    }
+
+    /// Sufficient buffer size for an encrypted message using the given aead algorithm
+    #[cfg(feature = "interface")]
+    pub fn aead_decrypt_output_size(self, alg: Aead, ciphertext_len: usize) -> Result<usize> {
```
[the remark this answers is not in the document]
sbailey-arm: I thought about doing that but I think I remember having this sort of conversation a while for a different helper method and it was decided that sticking it on `Attributes` was the most appropriate option at the time.

The PSA spec version does require `key_type` that is on `Attributes` and not on `Aead`, which is another reason I put it on `Attributes` (so it doesn't have to be moved in future to keep it in line with the existing methods).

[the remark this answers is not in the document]
ionut-arm: Ok, that makes sense if there's stuff from `Attributes` needed (or, will be)!

PR #56, opened 2020-08-18
Add Rust native key material types

joechrisellis:


joechrisellis:
cc: @ionut-arm 🙂

hug-dev:
I think the [`RsaPublicKey`](https://docs.rs/picky-asn1-x509/0.1.0/picky_asn1_x509/subject_public_key_info/struct.RSAPublicKey.html) type exists in `picky_asn1_x509`, the [`RsaPrivateKey`](https://docs.rs/picky-asn1-x509/0.1.0/picky_asn1_x509/private_key_info/struct.RSAPrivateKey.html) type exists but seems to be "less formatted".
Do we need to implement those specific types ourselves? For `RsaPrivateKey`, we could probably open an issue in picky to see if we could add more formatting?

joechrisellis:
@hug-dev you're right -- [`RsaPublicKey`](https://docs.rs/picky-asn1-x509/0.1.0/src/picky_asn1_x509/subject_public_key_info.rs.html#13-16) is defined as:

```
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
pub struct RSAPublicKey {
    pub modulus: IntegerAsn1,         // n
    pub public_exponent: IntegerAsn1, // e
}
```

Which is fine, but [`RsaPrivateKey`](https://docs.rs/picky-asn1-x509/0.1.0/src/picky_asn1_x509/private_key_info.rs.html#141-168) is defined as:

```
#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
pub struct RSAPrivateKey(Asn1SequenceOf<IntegerAsn1>);

impl RSAPrivateKey {
    pub fn modulus(&self) -> &IntegerAsn1 {
        &(self.0).0[1]
    }

    pub fn public_exponent(&self) -> &IntegerAsn1 {
        &(self.0).0[2]
    }

    pub fn private_exponent(&self) -> &IntegerAsn1 {
        &(self.0).0[3]
    }

    pub fn primes(&self) -> &[IntegerAsn1] {
        &(self.0).0[4..]
    }

    pub fn into_public_components(self) -> (IntegerAsn1, IntegerAsn1) {
        let mut iter = (self.0).0.into_iter();
        iter.next().expect("should not panic");
        (
            iter.next().expect("should not panic"),
            iter.next().expect("should not panic"),
        )
    }
}
```

Which AFAICT doesn't conform to the PSA standard, where the definition of the private key is:

```
RSAPrivateKey ::= SEQUENCE {
    version INTEGER, // must be 0
    modulus: INTEGER, // n
    public_exponent: INTEGER, // e
    private_exponent: INTEGER, // d
    prime_1: INTEGER, // p
    prime_2: INTEGER, // q
    exponent_1: INTEGER, // d mod (p - 1)
    exponent_2: INTEGER, // d mod (q - 1)
    coefficient: INTEGER, // (inverse of q) mod p
}
```

The picky `RsaPrivateKey` treats index 4 onwards as 'primes', and appears to omit (?) `exponent_1`, `exponent_2`, and `coefficient`, which is the main reason why we have our own type for that here. I also figured that if we're making our own private key type, we might as well have our own public key type, too, just for consistency's sake.

> For RsaPrivateKey, we could probably open an issue in picky to see if we could add more formatting?

We certainly could open an issue in picky if you would prefer to do this? 😄

hug-dev:
> The picky RsaPrivateKey treats index 4 onwards as 'primes', and appears to omit (?) exponent_1, exponent_2, and coefficient,

I believe that they might have done that because they RsaPrivateKey is just a vector and maybe the fields after `prime_2` are not defined/needed maybe?

I think it would definitely make sense to post an issue in [`picky`](https://github.com/Devolutions/picky-rs) to ask 👍

joechrisellis:
Closing in favour of (this PR)[https://github.com/Devolutions/picky-rs/pull/48]. The `picky` team were very receptive to our proposal, so there's no need for this any more. 😄

PR #58, opened 2020-08-28
Changed key derivation convertor to const function

sbailey-arm:
Rust v1.46 has been released which has support for match statements in a const function, so I've updated this method to be const.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,7 +7,13 @@ jobs:
     name: All tests
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):

sbailey-arm:
hmm, how do you update the version of Rust that the CI uses?

hug-dev:
I don't think they will upgrade the Rust version in their virtualised environment any time soon 😢 
However, you might want to use a GitHub Action that contain a more recent version of Rust: https://github.com/actions-rs/toolchain

hug-dev (approved):

PR #59, opened 2020-08-28
Fix indentation of the workflow

hug-dev:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,10 +10,10 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install latest Rust
         uses: actions-rs/toolchain@v1
-          with:
-            profile: minimal
-            override: true
-            toolchain: stable
+        with:
+          profile: minimal
+          override: true
+          toolchain: stable
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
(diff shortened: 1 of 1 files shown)


sbailey-arm:
Should we specify a version of Rust to use, or is sticking to the latest OK?

hug-dev:
Rust is stable, but I think sticking to a specific version is better in case new lints that might break CI are added. In my opinion, it is better to increase the version manually and fix all the problems in the same PR.

.github/workflows/ci.yml
```diff
@@ -10,10 +10,10 @@ jobs:
...
       - name: Install latest Rust
         uses: actions-rs/toolchain@v1
-          with:
-            profile: minimal
-            override: true
-            toolchain: stable
+        with:
+          profile: minimal
+          override: true
+          toolchain: stable
```
sbailey-arm, line 16: `toolchain: 1.46.0`

sbailey-arm (changes requested):
>Rust is stable, but I think sticking to a specific version is better in case new lints that might break CI are added. In my opinion, it is better to increase the version manually and fix all the problems in the same PR.
Agreed

.github/workflows/ci.yml
sbailey-arm, line 16:
```suggestion
          toolchain: 1.46.0
```

sbailey-arm (changes requested):

ionut-arm (approved):

sbailey-arm (approved):

PR #65, opened 2020-09-18
Remove const for backwards compatibility

### LINKED ISSUE #64: Delete const from pub const fn key_derivation
I can't build psa-crypto with version 1.45.2 of Rust because:
```
`match` is not allowed in a `const fn`
```

I can build it if I delete the `const` from `pub const fn key_derivation` here:

https://github.com/parallaxsecond/rust-psa-crypto/blob/master/psa-crypto/src/types/key_derivation.rs#L109

I don't think the `const` currently does anything useful so probably this change would be generally helpful.

hug-dev:
Close #64


egrimley-arm (approved):
Looks good to me!

ionut-arm (approved):

PR #68, opened 2020-12-08
Remove the bindgen dependency if not needed

hug-dev:
When compiling without no features, just the abstractred types are in, without any dependency on the bindgen generated types. Removing the dependency of bindgen makes the compilation of this crate faster and easier in those cases.


ionut-arm (approved):
Looks good!

PR #70, opened 2021-01-06
Fix clippy lint and disable Travis

ionut-arm:


hug-dev (approved):
Thanks for that!

PR #71, opened 2021-01-14
Bump bindgen and crate versions

ionut-arm:

### DIFF
--- a/psa-crypto-sys/Cargo.toml
+++ b/psa-crypto-sys/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "psa-crypto-sys"
-version = "0.6.1"
+version = "0.7.0"
 authors = ["Edmund Grimley Evans <edmund.grimley-evans@arm.com>",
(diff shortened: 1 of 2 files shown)


hug-dev:
Is that only in order to be able to increase the version of the `tss-esapi` crate in Parsec? Would it be possible to increase the `bindgen` version here but use a `git` dependency of this crate wherever needed?

ionut-arm:
How would we be able to publish any dependents, though? I think the whole dependency tree has to be in crates.io before you can publish

hug-dev:
> How would we be able to publish any dependents, though? I think the whole dependency tree has to be in crates.io before you can publish something

Yes, I see if we want to publish `parsec` as a crate on crates.io we will have to do that. Let's see what we decide [for this](https://cloud-native.slack.com/archives/C01EARH2ZB3/p1610526698198400)

ionut-arm:
Updated - we only need to make the service buildable :]

hug-dev (approved):

PR #73, opened 2021-02-11
Add method to retrieve hash from AsymSign

ionut-arm:
This commit adds a method to `AsymmetricSignature` to retrieve the hash
algorithm it is restricted to.

### DIFF
--- a/psa-crypto/src/types/algorithm.rs
+++ b/psa-crypto/src/types/algorithm.rs
@@ -437,6 +437,19 @@ impl AsymmetricSignature {
             }
         }
     }
+
+    /// Retrieves the specific hash that the `AsymmetricSignature` algorithm is restricted to.
+    ///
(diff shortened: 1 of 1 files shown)


psa-crypto/src/types/algorithm.rs
```diff
@@ -437,6 +437,19 @@ impl AsymmetricSignature {
             }
         }
     }
+
+    /// Retrieves the specific hash that the `AsymmetricSignature` algorithm is restricted to.
+    pub fn hash(self) -> SignHash {
```
hug-dev, line 442: I would personally prefer returning `Option<SignHash>` and return `None` for `EcdsaAny` and `RsaPkcs1v15SignRaw` so that all different cases are covered for this method. Maybe differentiating between those cases will be useful in the future?

hug-dev (approved):
Thanks!

PR #75, opened 2021-02-19
Implement Display on Algorithm

hug-dev:
Also fixes some clippy/deprecated errors.

### DIFF
--- a/psa-crypto/src/types/key.rs
+++ b/psa-crypto/src/types/key.rs
@@ -14,6 +14,7 @@ use crate::types::status::Status;
 use crate::types::status::{Error, Result};
 #[cfg(feature = "interface")]
 use core::convert::{TryFrom, TryInto};
+use core::fmt;
 use log::error;
 pub use psa_crypto_sys::{self, psa_key_id_t, PSA_KEY_ID_USER_MAX, PSA_KEY_ID_USER_MIN};
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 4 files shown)


psa-crypto/src/types/key.rs
```diff
@@ -541,10 +542,40 @@ pub enum Type {
...
+            Type::RawData => write!(f, "Raw data"),
+            Type::Hmac => write!(f, "HMAC key"),
+            Type::Derive => write!(f, "Derivation key"),
+            Type::Aes => write!(f, "Key for an algorithm based on AES"),
+            Type::Des => write!(f, "Key for an algorithm based on DES or 3DES"),
+            Type::Camellia => write!(f, "Key for an algorithm based on Camellia"),
+            Type::Arc4 => write!(f, "Key for the RC4 stream cipher"),
+            Type::Chacha20 => write!(
+                f,
+                "Key for the ChaCha20 stream cipher or the Chacha20-Poly1305 AEAD algorithm"
```
ionut-arm, line 557: Is this how it's defined in the spec? Seems sensible to just write it like the ones above `Key for an algorithm based on ChaCha20`

ionut-arm (approved):

psa-crypto/src/types/key.rs
hug-dev: Yees that's directly taken from the spec :/ https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_key_attributes.html#chacha20-type

ionut-arm: Hmm, though you did "sanatize" the ones above, they're not just copy-pasted

Issue #82, opened 2021-04-14
Fix the UsageFlags structure


hug-dev:
We added a clarification in the book in https://github.com/parallaxsecond/parsec-book/pull/110 regarding `sign/verify_hash/message`.

The following part is specifically important about `sign/verify_hash`:

> This flag automatically sets sign_message: if an application sets the flag sign_hash when creating a key, then the key always has the permissions conveyed by sign_message. For a key pair, this concerns the private key.

I see multiple ways to fix this:

1. Make the fields of `UsageFlags` private (or just the `sign`/`verify` ones?) with setters and getters on the different usages. `set_sign_hash` would set both `sign_hash` and `sign_message` to true.
2. Just add `verify/sign_message` methods that return `true` if either `sign/verify_message` or `sign/verify_hash` are `true`.

I would prefer 1 I think but that is a breaking change. That's probably fine. 2 would make the behaviour correct if you use the new method but we would have to check that we use that everywhere. It's not breaking though.

PR #83, opened 2021-04-28
Add PsaSignMessage and PsaVerifyMessage usage flags functions

akazimierskigl:
New functions for checking usage flags for PsaSignMessage and PsaVerifyMessage.

### DIFF
--- a/psa-crypto/src/types/key.rs
+++ b/psa-crypto/src/types/key.rs
@@ -874,15 +874,16 @@ impl From<UsageFlags> for psa_crypto_sys::psa_key_usage_t {
         if flags.decrypt {
             usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_DECRYPT;
         }
-        if flags.sign_message {
-            usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_MESSAGE;
-        }
+        //TODO: not yet implemented in Mbed Crypto, uncomment when added
+        //if flags.sign_message {
+        //  usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_MESSAGE;
+        //}
         if flags.sign_hash {
             usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_HASH;
         }
-        if flags.verify_message {
-            usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_VERIFY_MESSAGE;
-        }
+        //if flags.verify_message {
+        //  usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_VERIFY_MESSAGE;
+        //}
         if flags.verify_hash {
             usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_VERIFY_HASH;
         }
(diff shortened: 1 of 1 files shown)


ionut-arm (approved):
Thanks! This looks alright - are you planning to add the actual operations in a separate PR?

psa-crypto/src/types/key.rs
```diff
@@ -844,16 +874,15 @@ impl From<UsageFlags> for psa_crypto_sys::psa_key_usage_t {
...
-        //usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_MESSAGE;
-        //}
+        if flags.sign_message {
+            usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_MESSAGE;
+        }
         if flags.sign_hash {
             usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_SIGN_HASH;
         }
-        //if flags.verify_message {
-        //usage_flags |= psa_crypto_sys::PSA_KEY_USAGE_VERIFY_MESSAGE;
```
hug-dev, line 855: That was commented because the Mbed TLS version used by this crate does not support it, that's why the tests fail I think 😢
Very topical as this PR (https://github.com/ARMmbed/mbedtls/pull/4357) seems to add it! Meanwhile I think this is fine to keep this commented.
The whole permissions around sign/very hash/message will need to be revamped as part of #82 anyway

psa-crypto/src/types/key.rs
```diff
@@ -113,6 +113,36 @@ impl Attributes {
         }
     }
 
+    /// Check if a key has permission to sign a message
+    pub fn is_message_signable(self) -> bool {
+        self.policy.usage_flags.sign_hash & self.policy.usage_flags.sign_message
```
hug-dev, line 118: Shouldn't it be `|` here instead of `&`?

From the [`sign_hash`](https://parallaxsecond.github.io/parsec-book/parsec_client/operations/psa_key_attributes.html#sign-hash-usage) explanation:

> This flag automatically sets sign_message: if an application sets the flag sign_hash when creating a key, then the key always has the permissions conveyed by sign_message. For a key pair, this concerns the private key.

It means that a key can either have (ideally, this is currently not implemented, will be in #82):
1. `sign_hash` and `sign_message` to `false`
2. `sign_hash` and `sign_message` to `true`
3. `sign_hash` to `false` and `sign_message` to `true`

`&` would mean that 3 is not possible?

And same for `verify`. Not 100% sure so feel free to correct me!

ionut-arm: Yeah, that's true, I mis-remembered it the other way around - if you have `sign_message` you should also have `sign_hash` (since you're hashing the message and signing that...).

akazimierskigl: So it is possible scenario for 3. to exist? If yes then it make sense to change it to `|`.
As Ionut metioned and from point of view of cryptoauthlib you always need to be able to sign hash to sign message.

hug-dev: > So it is possible for scenario 3. to exist?

From the PSA Crypto API, yes it is! The scenario `sign_hash` to `true` and `sign_message` to `false` is not possible.

hug-dev (approved):
Thank you!

PR #86, opened 2021-07-07
Made the fields of UsageFlags private, Fix #82

### LINKED ISSUE #82: Fix the UsageFlags structure

Kakemone:
Made the fields of UsageFlags private and added a public setter and a public getter for each field.

Signed-off-by: Sam Davis <sam.davis@arm.com>


psa-crypto/tests/aead.rs
```diff
@@ -26,15 +26,15 @@ const ENCRYPTED_DATA: [u8; 40] = [
 #[test]
 fn aead_encrypt_aes_ccm() {
     let alg = Aead::AeadWithDefaultLengthTag(AeadWithDefaultLengthTag::Ccm);
+    let mut usage_flags: UsageFlags = Default::default();
+    dbg!(usage_flags.set_encrypt(true));
```
hug-dev, line 30: Could you please remove the `dbg!` macros (here and in other files)? So that other people that test locally don't have them printing something

psa-crypto/tests/mod.rs
```diff
@@ -12,23 +12,22 @@ mod key_agreement;
 
 #[test]
 fn generate_integration_test() {
+    let mut usage_flags: UsageFlags = Default::default();
+    usage_flags.set_sign_hash(false);
+    usage_flags.set_verify_hash(false);
+    usage_flags.set_verify_message(true);
```
hug-dev, line 18: everything should be set to `false` by default so only `verify_message` is necessary here! But it's fine, no need to change

psa-crypto/src/types/key.rs
```diff
@@ -770,25 +747,110 @@ pub struct Policy {
...
+    sign_hash: bool,
     /// Permission to verify a message hash with the key.
-    pub verify_hash: bool,
+    verify_hash: bool,
     /// Permission to derive other keys from this key.
-    pub derive: bool,
+    derive: bool,
+}
+
+impl UsageFlags {
```
hug-dev, line 771: I have two suggestions/simplifications for the setters/getters which are up for discussion:
- since everything is set to a secure default (`false`), I think one would only use `set_...(true)`, right? Because of that, maybe we could have the `set_...` methods not taking a `bool` parameter but setting their flag to `true`?
- I have seen this before in other code, and maybe we are doing it somewhere, where the getters do not have the `get_` prefix. For example `export` instead of `get_export` would return the `export` flag. Would that be ok?

hug-dev (commented):
It looks bigger than what I thought, thanks a lot for updating it all 👌 
Looks good to me, I wrote two suggestions up for debate.

ionut-arm (commented):
Looks good overall, a few comments on the getters!

ionut-arm (approved):
🚀

hug-dev (approved):
Thanks for the changes!

Issue #85, opened 2021-07-05
Compile issue with gcc 11


akuster:
Openembedded is now using gcc 11 and I am seeing the following build error.

parsec-service/0.7.0-r0/cargo_home/bitbake/psa-crypto-sys-0.8.0/vendor/library/psa_crypto.c:4590:12: error: parameter 'output_size' set but not used [-Werror,-Wunused-but-set-parameter]
|       size_t output_size,
|              ^
|   1 error generated.

Looking @ Mbedlts, it appears the offending function has been removed via commit: 

commit 6d05173359fb53eb2a893e6834a221a4262233da
Author: Ronald Cron <ronald.cron@arm.com>
Date:   Thu Oct 1 14:10:20 2020 +0200

    psa: Add mbedtls_psa_cipher_xyz() APIs
    
    Signed-off-by: Ronald Cron <ronald.cron@arm.com>

I think there is a mismatch somewhere.

hug-dev:
Hey 👋!!
Thanks for the bug report.

Looks like this commit is part of [this PR](https://github.com/ARMmbed/mbedtls/pull/4229) which was not included in the latest [2.26.0 release](https://github.com/ARMmbed/mbedtls/releases/tag/v2.26.0) 😢 
We are currently using 2.25.0 in this repo and ideally we would only point to Mbed TLS release versions. We could cherry-pick a bug fix on top of our git submodule but `6d05173359fb53eb2a893e6834a221a4262233da` seems to be way bigger than just this bug fix so Im wondering if other things will break if we try just to apply that.

This looks like a Mbed TLS bug, is there anything reported on their side?

I am also wondering if you could remove the `-Wunused-but-set-parameter` flag by setting your own `CFLAGS` before compiling `psa-crypto` with GCC 11

hug-dev:
Mbed TLS 3.0.0 was just [released](https://github.com/ARMmbed/mbedtls/releases/tag/v3.0.0). If the fix is included there, we can just use that in this crate and it should work ™️

ionut-arm:
It is, indeed, included!

akuster:
Apologies for the slow response. I have tried  -Wunused-but-set-parameter but  it does not propagate down to psa-crypto in the context of how this package is included within the OE build env. Sorry I can not be much help.

hug-dev:
[...]

PR #88, opened 2021-07-13
Move to Mbed TLS version 3.0.0

hug-dev:
See #85 for context

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -57,10 +57,14 @@ cargo build --no-default-features --features no-std
 # Test dynamic linking
 git clone https://github.com/ARMmbed/mbedtls.git
 pushd mbedtls
-git checkout mbedtls-2.25.0
+git checkout v3.0.0
 ./scripts/config.py crypto
 SHARED=1 make
 popd
+
+# Clean before to only build the interface
+cargo clean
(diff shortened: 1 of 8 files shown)


hug-dev:
This is a breaking change since parameters of methods change. Won't be as easy as doing `cargo update` to integrate that everywhere.

ionut-arm (approved):
I was worried what effect this would have on the TS integration, but it doesn't look like that much changed. Thanks for putting this together!

hug-dev:
The ABI of Mbed Crypto changed from 2.25.0 and 3.0.0 which means that this crate will only be compatible with Mbed TLS 3.0.0 and later versions. This has impact for Parsec stability and for SE driver.

hug-dev:
It was decided to go on with this.

There was a discussion on [Mbed TLS mailing list](https://lists.trustedfirmware.org/pipermail/mbed-tls/2021-July/000409.html) to know more about the API/ABI stability of Mbed Crypto in Mbed TLS.
Until a PSA Crypto compliant version of Mbed Crypto is in an LTS version of Mbed TLS, we can't offer a stable dynamic linking process on this crate. Building from source and statically linking is fine for now.

PR #93, opened 2021-08-07
psa-crypto-sys: bump bindgen to 0.57.0

nullr0ute:
It builds fine with the newer version and other deps in the
parsec dependency chain already depend on 0.57.0.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


ionut-arm (approved):

hug-dev (approved):
Thank you! We will wait that everything is ok on your side before starting making new releases.
Hopefully, this should only be a patch release which means that we could potentially directly import it in Parsec.

PR #95, opened 2021-11-29
Add checks for keys sizes for psa_key_attributes_t convert

anta5010:
Signed-off-by: Anton Antonov <Anton.Antonov@arm.com>

### DIFF
--- a/psa-crypto/src/types/key.rs
+++ b/psa-crypto/src/types/key.rs
@@ -178,6 +178,13 @@ impl Attributes {
         }
     }
 
+    /// Check if can be converted into psa_key_attributes_t
+    #[cfg(feature = "interface")]
+    pub fn can_convert_into_psa(self) -> Result<()> {
+        let _ = psa_crypto_sys::psa_key_attributes_t::try_from(self)?;
+        Ok(())
+    }
+
     /// Check if the alg given for a cryptographic operation is permitted to be used with the key
     pub fn is_alg_permitted(self, alg: Algorithm) -> bool {
         match self.policy.permitted_algorithms {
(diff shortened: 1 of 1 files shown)


psa-crypto-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "psa-crypto-sys"
-version = "0.9.0"
+version = "0.9.1"
```
ionut-arm, line 3: Do we need to bump this? Is it for the `build.rs` change?

psa-crypto/src/types/key.rs
```diff
@@ -897,12 +903,106 @@ impl TryFrom<Attributes> for psa_crypto_sys::psa_key_attributes_t {
...
+                    },
+                    _ => {
+                        // We don't (yet?) implement checks for other curve families
+                        Ok(attributes.bits)
+                    }
+                }
+            }
+            _ => {
+                // TO-DO We don't (yet?) implement checks for other types
+                Ok(attributes.bits)
```
ionut-arm, line 1000: Doesn't this break any tests that involve symmetric primitives, like AEAD?

anta5010: Why would it? Functionality is the same as before for all types we haven't implemented checks in this function.

psa-crypto-sys/Cargo.toml
anta5010: This new functionality is required for can-do-crypto for TS and mbed-crypto providers. So, I was planning to bump version here and in parsec and parsec-interface cargo.toml. If I can avoid the bump it would make my life easier.

psa-crypto/src/types/key.rs
ionut-arm: Ah, yes, of course, apologies, for some reason I only read the "TODO" line and just assumed it returns an error 🤦🏻

psa-crypto-sys/Cargo.toml
ionut-arm: I don't think you need to bump the `-sys` crate unless you want to release something specific. Might be good to release that change from late last week, though!

anta5010: removed the bump from `-sys`

ionut-arm (approved):

PR #97, opened 2021-11-30
Update CHANGELOG

anta5010:
Signed-off-by: Anton Antonov <Anton.Antonov@arm.com>


ionut-arm:
I think we need to find a better way of doing this.

Could you copy just the part corresponding to the latest release and put it on top of the old changelog file? Though some of the changes this time seem legitimate, but if you rerun it a few times in a row you'll notice it keeps moving some issues/bugfixes etc. from one release to another (as in, old releases). Would be nice if that didn't happen, maybe there's a CLI flag that we can pass in the future to do this.

anta5010:
> Could you copy just the part corresponding to the latest release and put it on top of the old changelog file?

done

ionut-arm (approved):

PR #99, opened 2022-01-25
build.rs: Build Mbed TLS with ENABLE_PROGRAMS=OFF, ENABLE_TESTING=OFF.

egrimley-arm:
In building psa-crypto-sys we never execute the Mbed TLS tests, so
there is little point in building them. By not building them we save
CPU cycles and disc space and we make these crates a little more
portable as the tests use C library facilities that are not required
by the library.


ionut-arm (approved):
Thanks! Out of curiosity, what does `ENABLE_PROGRAMS` do?

hug-dev (approved):
Good find!

egrimley-arm:
> Thanks! Out of curiosity, what does `ENABLE_PROGRAMS` do?

Example programs in the `programs` directory, I think.

PR #100, opened 2022-01-25
build.rs: Use generated config file with shim library headers.

egrimley-arm:
This probably makes no practical difference with the current version
of Mbed TLS and the current configuration but it might matter in the
future so this change could make psa-crypto-sys more maintainable.

At the top of "shim.h" we include the config file in the same way that
Mbed TLS .c files do, and in "build.rs" we add the output directory to
the include path and define MBEDTLS_CONFIG_FILE.

To make this work without the "operations" feature the code that
configures Mbed TLS is moved out of compile_mbed_crypto into a new
function, configure_mbed_crypto, that is called from script_interface
and script_operations.

### DIFF
--- a/psa-crypto-sys/build.rs
+++ b/psa-crypto-sys/build.rs
@@ -45,14 +45,53 @@ fn main() -> std::io::Result<()> {
 mod common {
     use std::env;
     use std::io::{Error, ErrorKind, Result};
-    use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


ionut-arm (approved):
Thanks, LGTM!

ionut-arm (approved):

anta5010 (approved):

PR #104, opened 2022-03-17
Add unauthenticated cypher operations in psa-crypto and psa-crypto-sys.

egrimley-arm:


psa-crypto/src/operations/cipher.rs
```diff
@@ -2,3 +2,146 @@
...
+/// let iv = vec![0; 16];
+/// let mut encrypted_message = vec![0; MESSAGE.len()];
+///
+/// let size = encrypt(my_key, alg, &MESSAGE, &iv, &mut encrypted_message).unwrap();
+/// ```
+pub fn encrypt(
+    key_id: Id,
+    alg: Cipher,
+    plaintext: &[u8],
+    iv: &[u8],
```
ionut-arm, line 101: Could you make the `iv` field optional for encryption, and call `psa_cipher_generate_iv` if it's not provided?

psa-crypto/src/operations/cipher.rs
```diff
@@ -2,3 +2,146 @@
...
+        } else {
+            psa_crypto_sys::psa_cipher_decrypt_setup
+        })(&mut operation, key_id.0, alg.into())
+    })
+    .to_result()?;
+
+    Status::from(unsafe {
+        psa_crypto_sys::psa_cipher_set_iv(&mut operation, iv.as_ptr(), iv.len())
+    })
+    .to_result()?;
```
ionut-arm, line 36: I think for every call following a successful `psa_cipher_operation_init`, if the operation fails you need to call `psa_cipher_abort` before returning.

ionut-arm (commented):
Thanks, looks good overall!

psa-crypto/src/operations/cipher.rs
egrimley-arm: > Could you make the `iv` field optional for encryption, and call `psa_cipher_generate_iv` if it's not provided?

I don't think that makes sense with any of these algorithms because the IV isn't included in the encrypted message so a message generated with a random IV that isn't revealed to the caller has no feasible use.

ionut-arm: Ah, I was under the impression the generated IV would be prepended to the ciphertext. Hmmm, I would still prefer if the IV could be left to the implementation to handle, especially since it's the recommended approach in the spec. Will try to think of a way

ionut-arm: Hmm, perhaps we can leave as is for now and add another method, something like `encrypt_with_new_iv`, when there's a need.

psa-crypto/src/operations/cipher.rs
```diff
@@ -2,3 +2,157 @@
...
+    }
+
+    let mut output_length_finish = 0;
+    Status::from(unsafe {
+        psa_crypto_sys::psa_cipher_finish(
+            &mut operation,
+            ciphertext.as_mut_ptr().add(output_length),
+            ciphertext.len() - output_length,
+            &mut output_length_finish,
+        )
```
ionut-arm, line 69: I think this needs to be covered by the `match` above as well

ionut-arm (approved):
👍🏻

psa-crypto/src/operations/cipher.rs
egrimley-arm: You're absolutely right. I failed to notice the word "successful" in this bit in `crypto.h`:
```
 * After a successful call to psa_cipher_encrypt_setup(), the application must
 * eventually terminate the operation. The following events terminate an
 * operation:
 * - A successful call to psa_cipher_finish().
 * - A call to psa_cipher_abort().
```

anta5010 (approved):
LGFM as far as I understand the code

PR #107, opened 2022-08-31
Bump zeroize version

gowthamsk-arm:
Updated to 1.4.3 not 1.5.* as the rsa crate has dependencies issue

Relates to #602


ionut-arm (approved):
Have you tried to compile it with the Parsec service MSRV?

gowthamsk-arm:
I've compiled this crate with v1.53.0 using `cargo +1.53.0 build` command and it builds successfully. Can you please tell me if there is there any other check I need to do?

ionut-arm:
> Can you please tell me if there is there any other check I need to do?

No, my question was only related to that comment in the `Cargo.toml` file that you removed, wanted to make sure that was tested and we don't have to worry about it anymore!

PR #113, opened 2023-08-01
Disable default-features on serde

ithinuel:
Serde enables by default its std features, we do not want that.


ionut-arm:
Seems we're having the same problem as we did with `tss-esapi` - compiler too old for the `log` crate. Do we want to bump MSRV?

egrimley-arm:
You could remove the change to `ci.sh` and rebase this onto `main` now.

ithinuel:
Updated. Hopefully it'll pass CI 🤞

PR #115, opened 2023-08-04
Add a prefix to the library names and to every global symbol

egrimley-arm:
A collision with another crate, even another crate that uses the same C library, should now be very unlikely, so the `links` line in `Cargo.toml` is removed. It should also be possible for a project to use multiple versions of this crate.

The libraries are now called `libpsa_crypto_0_10_0_shim.a` and `libpsa_crypto_0_10_0_mbedcrypto.a` and there is a new directory for them (`.../out/llib`) which does not contain any other libraries. All the globally defined symbols have the prefix `psa_crypto_0_10_0_`.

This relies on `bindgen::callbacks::ParseCallbacks::generated_name_override`, which was added in `bindgen` 0.64.0, which requires Rust 1.60.0. Currently we want `psa-crypto` to work with Rust 1.58.1. We could perhaps make this prefixing optional, enabled by a feature, but I don't think there's any way to make the `links` field in `Cargo.toml` conditional on a feature so perhaps we will just have to wait till people are happy to move to `rust-version = "1.60.0"`.

I've done this on top of #116, which tests with two versions of Rust, and I've updated the minimal version from 1.58.1 to 1.60.0 in a separate commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,15 +3,15 @@ name: Continuous Integration
 on: [push, pull_request]
 
 jobs:
-  build-1-58-1:
-    name: "All tests: rustc 1.58.1"
+  build-1-60-0:
+    name: "All tests: rustc 1.60.0"
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - name: Install Rust 1.58.1
+      - name: Install Rust 1.60.0
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: 1.58.1
+          toolchain: 1.60.0
           rustflags: ""
       - name: Set up Python
         uses: actions/setup-python@v4
--- a/psa-crypto-sys/Cargo.toml
+++ b/psa-crypto-sys/Cargo.toml
@@ -17,6 +17,9 @@ cc = "1.0.59"
 cmake = "0.1.44"
 regex = "1.9.1"
 walkdir = "2.3.1"
+# Later versions of "memchr" and "which" (indirectly) need rustc >1.60.0:
+memchr = "=2.6.2"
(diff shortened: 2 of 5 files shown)


ionut-arm (commented):
Amazing solution! Makes me think whether we should do this for other crates.

The code looks good, the one request I have is that you add a short notice about this in the psa-crypto-sys README.

egrimley-arm:
> The code looks good, the one request I have is that you add a short notice about this in the psa-crypto-sys README.

Done. Tell me if you think of any other comments/documentation that might be useful.

ionut-arm (approved):
Thanks!

gowthamsk-arm:
Is this still in draft state or can I review it? :)

egrimley-arm:
> Is this still in draft state or can I review it? :)

I've left this as "draft" because it requires (and includes as separate commits) updating the minimal Rust version from 1.58.1 to 1.60.0. I think someone said that there's a good reason for the minimal Rust version being 1.58.1, but it would be nice to document here what that reason is so that we can know when that reason has gone away.

But that's the only reason I've left this as "draft".

tgonzalezorlandoarm:
Hello @egrimley-arm could you resolve the conflicts? IMO this is ready for final review and merge

egrimley-arm:
It looks like an update to some crate we depend on has made `rustc 1.60` no longer work. Perhaps I could make it work with 1.60 by adding some constraint to the `Cargo.toml`, but as far as I know there's nothing special about 1.60 so perhaps we should just increase the Rust version we test with.

tgonzalezorlandoarm:
@egrimley-arm Yes, if possible, could you use 1.66.0 instead? Thanks

tgonzalezorlandoarm (approved):

PR #117, opened 2023-08-22
Update GitHub workflows to use newer actions.

egrimley-arm:
These are deprecated:
actions/checkout@v2, actions-rs/toolchain@v1, actions/setup-python@v2.

You can see a warning about it here, for example:
https://github.com/parallaxsecond/rust-psa-crypto/actions/runs/5937148892


gowthamsk-arm (approved):
LGTM. Thanks :)

ionut-arm (approved):

PR #127, opened 2023-10-16
Add prefix feature

gowthamsk-arm:
The prefix feature adds a psa_crypto_(version) prefix to the library names and all the symbols. By default it is disabled.
    

gowthamsk-arm:
With the prefix feature enabled, there are some issues related to cross-compiling. This will be fixed in another PR. Having it as a feature and enabling only when needed would be better.

ionut-arm (approved):
👍🏻

tgonzalezorlandoarm:
Should we add a test for the non-prefix option?

tgonzalezorlandoarm (approved):
LGTM

gowthamsk-arm:
What is the general approach to add features:

1. "no-sym-prefix" enable it by default and users need to explicitly disable
2. "prefix" disabled by default and users need to explicitly enable prefixing when they want

I'm more inclined towards option 2 as having lots of features enabled by default might not be convenient. Any thoughts?

psa-crypto-sys/build.rs
```diff
@@ -35,6 +35,16 @@
 #![allow(clippy::multiple_crate_versions)]
 
 fn main() -> std::io::Result<()> {
+    // If the prefix feature is not enabled then set the "CARGO_PKG_LINKS"
+    // parameter to mbedcrypto to avoid any duplicate symbols from any other
```
egrimley-arm, line 39: Should this be `mbedtls` instead of `mbedcrypto` so that we can detect an attempt to combine with `mbedtls-sys-auto` (https://github.com/fortanix/rust-mbedtls/blob/master/mbedtls-sys/Cargo.toml#L14)?

tgonzalezorlandoarm:
> What is the general approach to add features:
> 
> 1. "no-sym-prefix" enable it by default and users need to explicitly disable
> 2. "prefix" disabled by default and users need to explicitly enable prefixing when they want
> 
> I'm more inclined towards option 2 as having lots of features enabled by default might not be convenient. Any thoughts?

I agree with option 2, I would say that the criteria is not to add any features by "surprise", so if you disable the defaults, you have to explicitly enable each feature that you want enabled. This may result in a longer Cargo.toml line, but it avoid the surprise factor of bringing along a crate that you weren't expecting to bring

psa-crypto-sys/build.rs
gowthamsk-arm: I'm worried changing this would affect if there is another crate that links to `libmbedcrypto`. The user in this case can use the `prefix` feature to avoid any collisions instead of changing `mbedcrypto`?

egrimley-arm: In fact I've also suggested that `mbedtls-sys-auto` change its `links` line: https://github.com/fortanix/rust-mbedtls/issues/289
If both crates changed, then we'd be back to where we started! The `links` line is a very crude mechanism for detecting collisions.

gowthamsk-arm: Since you have already raised an issue there, I think we can stick to `mbedcrypto` for now.

ionut-arm:
I agree with option 2 as well - as long as we've not had the renaming as a default in the previous release (?). Having more features to enable isn't a problem, especially if they can have drastic effects on how the crate works for dependents.

gowthamsk-arm:
Initially, there was no prefixing, but then `v0.11.0` made prefixing as default (without any feature). Now we are making prefix non-default.

gowthamsk-arm:
Provided `v0.12.0` release will be a quick release after `v0.11.0` should we be concerned about users having issues with it? :)

ionut-arm:
Hmm, ok, I guess as long as it's documented somewhere, this should be ok.

gowthamsk-arm:
Makes sense. Will do that:)

Issue #131, opened 2023-11-24
psa_crypto_sys is not no_std


chrysn:
Starting to look into how this could be used on RIOT OS (that since its 2023.10 release has PSA support on its C side), I've noticed that psa_crypto_sys is no_std.

Is that an oversight (if so, it may be convenient to test that things are buildable on no_std, eg. by including a build with `--target `, similar to the CI commits of https://github.com/paritytech/wasmi/pull/387); if not: is there a plan for how this is to be used on embedded? (Indeed it might be that for RIOT's purposes the psa_crypto_sys crate is patched to go through a different header discovery route, though I'd hope that this can be done upstream by just providing an environment variable).

hug-dev:
Hello! Hopefully after #145 this crate will be no_std! Happy to follow your integration in RIOT OS :) Hope it's not too late!

PR #133, opened 2023-11-28
Add no-std and baremetal features

geonnave:
This is an ongoing attempt at addressing #131. Heavily based on [this fork](https://github.com/malishav/rust-psa-crypto/tree/baremetal). cc @chrysn @malishav


chrysn:
Adding a baremetal and a no_std feature is contrary to the addititvity of Cargo features. If there is an actual need for std (for #131 I don't think we need an extra feature, it can be no_std and no_alloc all the time), and it's on by default now, it should become a default feature "std", so that all dependencies that are themselves no_std would use this with no-default-features, but as soon as any dependent crate needs std, std is in.

geonnave:
Thanks for the feedback.

Before I spend more time on this, I would like to hear what the maintainers think of addressing this issue.

ionut-arm:
Hi!

Thanks for the patch, it looks good to me. In my opinion, making it `no_std` is a good step forward, but happy to hear other views too.

I wanted to ask whether it would be possible/make sense to enable it for `psa-crypto` too.

Also, the one thing I'd like added is some mention in the README about the new features. :)

geonnave:
Thank you @ionut-arm. 

Regarding `psa-crypto`, [it seems to already be `no-std` except when using the `std` feature](https://github.com/parallaxsecond/rust-psa-crypto/blob/main/psa-crypto/src/lib.rs#L12) (which I think is more like what @chrysn was proposing to have it working).

In addition, I realized I need some extra build flags to be passed to `mbedtls` when building this in `no-std` mode for an embedded target (nRF52840). Unfortunately, it requires patching mbedtls. I added a patch and the required logic in `build.rs`, but I am still not sure how general this solution is, i.e. will others wanting to build in baremetal configuration need this same patch? (cc @malishav)

PR #134, opened 2023-11-30
audit.toml: Add Marvin Attack from rsa crate to ignore list

gowthamsk-arm:
The `rsa` crate is only used for testing purpose and the security vulnerability has no impact on this crate.


.cargo/audit.toml
```diff
@@ -0,0 +1,4 @@
+[advisories]
+ignore = [
+    "RUSTSEC-2023-0071", # `rsa` Marvin Attack not concerning as `rsa` is only used for tests
```
tgonzalezorlandoarm, line 3: What happens if we suddenly use it elsewhere (not in tests) ? Can we ignore it only for certain files?

tgonzalezorlandoarm (changes requested):

.cargo/audit.toml
gowthamsk-arm: Yeah, it might happen in other repos. I'm unaware if there is an option to restrict audit ignores to specific files. Let me have a look.

tgonzalezorlandoarm (approved):
LGTM

PR #135, opened 2024-01-02
Mbedtls fixes 0.12

billatarm:
- ~Fix linking options against system provided mbedtls~ It was me using an ETOOOLD version of mbedtls. Just don't run configure on vendor'd mbedtls wqhen env vars set.
  - Probably add a CI test for this, something like rm vendor directory, set build flags and ensure build works. Maybe even leer into open files of the process using the bindings and ensure its loaded mbedtls.
- Deduplicate parts of build.rs

### DIFF
--- a/psa-crypto-sys/build.rs
+++ b/psa-crypto-sys/build.rs
@@ -57,6 +57,8 @@ fn main() -> std::io::Result<()> {
 
 #[cfg(any(feature = "interface", feature = "operations"))]
 mod common {
+    pub const CONFIG_FILE: &str = "custom_config.h";
+
     #[cfg(feature = "prefix")]
     use bindgen::callbacks::{ItemInfo, ParseCallbacks};
 
@@ -84,7 +86,7 @@ mod common {
         // Configure the MbedTLS build for making Mbed Crypto
         if !::std::process::Command::new(mbedtls_config)
             .arg("--write")
-            .arg(&(out_dir + "/config.h"))
+            .arg(&(out_dir + "/" + CONFIG_FILE))
             .arg("crypto")
             .status()
             .map_err(|_| Error::new(ErrorKind::Other, "configuring mbedtls failed"))?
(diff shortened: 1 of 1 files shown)


psa-crypto-sys/build.rs
```diff
@@ -250,10 +262,15 @@ mod operations {
             }
         }
 
+        // if we're compiling mbedtls, we're not using the system provided
+        // package
+        let mbed_conf = common::get_mbedtls_conf(false);
+
         // Build the MbedTLS libraries
+        //
```
billatarm, line 257: oops, let me drop this.

gowthamsk-arm:
Thanks for the PR @billatarm :)
There are some trivial CI failures. Can you fix them, please?

billatarm:
> Thanks for the PR @billatarm :) There are some trivial CI failures. Can you fix them, please?

weird I wonder why these didn't trigger locally.

billatarm:
> > Thanks for the PR @billatarm :) There are some trivial CI failures. Can you fix them, please?
> 
> weird I wonder why these didn't trigger locally.

ahh features

billatarm:
hrmm some weird memory issues now, I wonder if it's getting some wrong definitions for type sizes when building the shim. I think its my usage of MBEDTLS_CONFIG_FILE

billatarm:
> hrmm some weird memory issues now, I wonder if it's getting some wrong definitions for type sizes when building the shim. I think its my usage of MBEDTLS_CONFIG_FILE

Ahh fedora has an ancient mbedtls 2.28.5 and we need 3.0 minimum. Half the changes in this patch are because I didn't catch how old Fedora's package was. This greatly simplifies the patch.

gowthamsk-arm (dismissed):
LGTM.

psa-crypto-sys/build.rs
```diff
@@ -120,70 +122,74 @@ mod common {
...
-            .define("MBEDTLS_CONFIG_FILE", "<config.h>")
-            .include(include_dir)
-            .file("./src/c/shim.c")
-            .warnings(true)
-            .flag("-Werror")
-            .opt_level(2)
-            .cargo_metadata(metadata)
-            .try_compile(shimlib_name)
+        let mut cfg = cc::Build::new();
+        _ = cfg.include(&out_dir);
```
tgonzalezorlandoarm, line 180: nit: Why are we doing `_ = cfg.function()` instead of the `cfg.function1().function2()` pattern? Can we not leave it as it was?

psa-crypto-sys/build.rs
```diff
@@ -120,70 +122,74 @@ mod common {
...
-            .size_t_is_usize(true)
-            .parse_callbacks(Box::new(RenameCallbacks {}))
-            .generate()
-            .map_err(|_| {
-                Error::new(
-                    ErrorKind::Other,
-                    "Unable to generate bindings to mbed crypto",
-                )
-            })?;
+        {
```
tgonzalezorlandoarm, line 146: nit: Only one line, so maybe we can remove the curly braces?

tgonzalezorlandoarm (changes requested):
Two nits, LGTM apart from that

psa-crypto-sys/build.rs
williamcroberts: > nit: Why are we doing `_ = cfg.function()` instead of the `cfg.function1().function2()` pattern? Can we not leave it as it was?

I tried that, it's what I wanted to do, originally I couldn't find a way to make the borrow checker happy, but I just found a way. Here is was the error, change incoming.
```bash
error[E0716]: temporary value dropped while borrowed
   --> psa-crypto-sys/build.rs:179:23
```

billatarm: > nit: Why are we doing `_ = cfg.function()` instead of the `cfg.function1().function2()` pattern? Can we not leave it as it was?

I tried that, it's what I wanted to do, originally I couldn't find a way to make the borrow checker happy, but I just found a way. Here is was the error, change incoming.
```bash
error[E0716]: temporary value dropped while borrowed
   --> psa-crypto-sys/build.rs:179:23
```

psa-crypto-sys/build.rs
tgonzalezorlandoarm: Seeing the comparison I think the change you applied is in a different line that I intended. I was suggesting replacing
```
        #[cfg(feature = "prefix")]
        {
            shim_builder = shim_builder.parse_callbacks(Box::new(RenameCallbacks {}));
        }
```

with
```
        #[cfg(feature = "prefix")]
         shim_builder = shim_builder.parse_callbacks(Box::new(RenameCallbacks {}));
```

psa-crypto-sys/build.rs
tgonzalezorlandoarm: Thanks :)
There is an extra tab that is being flagged by the CI.

psa-crypto-sys/build.rs
billatarm: That moment you seek to the wrong line in vim haha

billatarm: @tgonzalezorlandoarm that seems to result in:
```bash
Compiling psa-crypto-sys v0.12.0 (/home/bill/workspace/rust-psa-crypto/psa-crypto-sys)
error[E0658]: attributes on expressions are experimental
   --> psa-crypto-sys/build.rs:145:9
    |
145 |         #[cfg(feature = "prefix")]
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #15701 <https://github.com/rust-lang/rust/issues/15701> for more information
    = help: add `#![feature(stmt_expr_attributes)]` to the crate attributes to enable

For more information about this error, try `rustc --explain E0658`.
```

tgonzalezorlandoarm: mmm okay, nevermind then. Thanks!

tgonzalezorlandoarm (approved):
LGTM

gowthamsk-arm (approved):

ionut-arm (approved):
Thank you, looks sensible to me!

PR #137, opened 2024-01-19
Update mbedtls to version 3.5.1

billatarm:
-  mbedtls: update to version 3.5.1
- build: support lib64 out directories

Could we get a release with this? (assuming no known bugs, worked on my systems :-p)

### DIFF
--- a/psa-crypto-sys/build.rs
+++ b/psa-crypto-sys/build.rs
@@ -308,6 +308,10 @@ mod operations {
             let mut mbed_lib_dir = compile_mbed_crypto()?;
             let mut mbed_include_dir = mbed_lib_dir.clone();
             mbed_lib_dir.push("lib");
+            if !mbed_lib_dir.as_path().exists() {
+                _ = mbed_lib_dir.pop();
+                mbed_lib_dir.push("lib64");
+            }
             mbed_include_dir.push("include");
 
             lib = mbed_lib_dir.to_str().unwrap().to_owned();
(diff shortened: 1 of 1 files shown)


nullr0ute:
Yes please, there's some CVEs between 3.0 and 3.5.1

psa-crypto-sys/build.rs
```diff
@@ -388,6 +397,7 @@ mod operations {
             let (from, _) = &lib;
             args.push(from.as_os_str());
         }
+        println!("BILL: args: {:?}", args);
```
gowthamsk-arm, line 400: Can this be dropped?

psa-crypto-sys/build.rs
```diff
@@ -357,6 +361,11 @@ mod operations {
             let mut mbed_lib_dir = compile_mbed_crypto()?;
             let mut mbed_include_dir = mbed_lib_dir.clone();
             mbed_lib_dir.push("lib");
+            if !mbed_lib_dir.as_path().exists() {
+                _ = mbed_lib_dir.pop();
+                mbed_lib_dir.push("lib64");
```
gowthamsk-arm, line 366: On my machine, it is just "lib". Is this a special case?

billatarm: > On my machine, it is just "lib". Is this a special case?

Yeah when I updated mbedtls on my Fedora 39 machine, it uses lib64 over lib.

psa-crypto-sys/build.rs
billatarm: Yeah whoops, I thought I dropped all of those :-p

gowthamsk-arm (approved):
LGTM. Thanks :)

tgonzalezorlandoarm (approved):
LGTM, Thanks!

PR #139, opened 2024-01-23
Use pkg config

billatarm:


psa-crypto-sys/build.rs
```diff
@@ -209,22 +276,17 @@ mod common {
...
-            Ok(())
-        } else {
-            Err(Error::new(
-                ErrorKind::Other,
-                "interface feature necessitates MBEDTLS_INCLUDE_DIR environment variable",
-            ))
-        }
+        let include_dir = common::get_external_mbedtls_include_only().unwrap();
+
+        // TODO: Does interface need the vendored mbedtls?
```
billatarm, line 285: @gowthamsk-arm is this needed? I don't think it's needed.

gowthamsk-arm: I don't think it's needed. It was added in this [patch](https://github.com/parallaxsecond/rust-psa-crypto/commit/c66aa62349de40ceb395c681f0f8abe126916707). 

@egrimley-arm, just to confirm, was there a particular reason (that we are missing) to configure the vendored medtls even for the interface feature?

psa-crypto-sys/build.rs
```diff
@@ -388,6 +448,7 @@ mod operations {
             let (from, _) = &lib;
             args.push(from.as_os_str());
         }
+        println!("BILL: args: {:?}", args);
```
tgonzalezorlandoarm, line 451: I think this line is extra?

psa-crypto-sys/build.rs
```diff
@@ -66,6 +66,73 @@ mod common {
...
+            return Some(Err(Error::new(
+                ErrorKind::Other,
+                "both environment variables MBEDTLS_LIB_DIR and MBEDTLS_INCLUDE_DIR need to be set for operations feature",
+            )));
+        }
+
+        if let (Ok(lib_dir), Ok(include_dir)) =
+            (env::var("MBEDTLS_LIB_DIR"), env::var("MBEDTLS_INCLUDE_DIR"))
+        {
+            println!("Found environment varibales, using external MbedTLS");
```
tgonzalezorlandoarm, line 81:
```suggestion
            println!("Found environment variables, using external MbedTLS");
```

psa-crypto-sys/build.rs
billatarm: Ha I thought I deleted that!

psa-crypto-sys/build.rs
```diff
@@ -66,6 +66,73 @@ mod common {
     use std::io::{Error, ErrorKind, Result};
     use std::path::{Path, PathBuf};
 
+    #[cfg(any(feature = "prefix", feature = "operations"))]
+    pub fn get_external_mbedtls() -> Option<Result<(String, String)>> {
+        if env::var("MBEDTLS_LIB_DIR").is_err() ^ env::var("MBEDTLS_INCLUDE_DIR").is_err() {
```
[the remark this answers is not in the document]
billatarm: I just copied the existing behavior found at https://github.com/parallaxsecond/rust-psa-crypto/blob/main/psa-crypto-sys/build.rs#L291. 

is_err will return True on an unset variable, so really what we want is if one returns success and the other is error, fail.

| LIB_DIR | INCLUDE_DIR | RESULT |
| -------- | -------------- | -------- |
|  Ok        | Ok                    | Use env vars |
| err        | Ok                    | Error |
| Ok.       | Err                    | Error |
| Err.       | Err                    | Skip and use vendored or pkg-config mbedtls |

That truth. table is XOR

psa-crypto-sys/build.rs
```diff
@@ -66,6 +66,73 @@ mod common {
...
+        }
+
+        // No env vars set and no discovered package through pkg-config
+        None
+    }
+
+    #[cfg(all(feature = "interface", not(feature = "operations")))]
+    pub fn get_external_mbedtls_include_only() -> Result<String> {
+        if let Ok(include_dir) = env::var("MBEDTLS_INCLUDE_DIR") {
+            println!("Found environment varibale, using external MbedTLS");
```
tgonzalezorlandoarm, line 112:
```suggestion
            println!("Found environment variable, using external MbedTLS");
```

psa-crypto-sys/build.rs
```diff
@@ -209,22 +276,17 @@ mod common {
...
-            common::generate_mbed_crypto_bindings(include_dir.clone(), false)?;
-            let _ = common::compile_shim_library(include_dir, true, false)?;
-            Ok(())
-        } else {
-            Err(Error::new(
-                ErrorKind::Other,
-                "interface feature necessitates MBEDTLS_INCLUDE_DIR environment variable",
-            ))
-        }
+        let include_dir = common::get_external_mbedtls_include_only().unwrap();
```
tgonzalezorlandoarm, line 283: We are not unwrapping in the rest of the function calls, should we not use `?` and handle the error afterwards as well?

tgonzalezorlandoarm (changes requested):

psa-crypto-sys/build.rs
billatarm: > We are not unwrapping in the rest of the function calls, should we not use `?` and handle the error afterwards as well?

That's not true, on current main their are 19 uses of unwrap():
  - `git grep unwrap build.rs | wc -l`

I mean we can use question mark operator here I'm not opposed, but it didn't seem like a recoverable error so panic would suffice and the output is pretty identical. But ill switch to `?`

With panic on unwrap:
```
  --- stderr
  thread 'main' panicked at psa-crypto-sys/build.rs:283:71:
  called `Result::unwrap()` on an `Err` value: Custom { kind: Other, error: "interface feature necessitates MBEDTLS_INCLUDE_DIR environment variable" }
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 ```
 
 Versus `?`:
 ```
  --- stderr
  Error: Custom { kind: Other, error: "interface feature necessitates MBEDTLS_INCLUDE_DIR environment variable" }
```

psa-crypto-sys/build.rs
billatarm: I dropped this in a separate patch just on the off chance it needs to be reverted.

billatarm: Digging into this more, this is not what you wanted as the interface feature was partially broken as well. I kept the behavior the same until now, as I noticed that it was borked. See the commit message.

psa-crypto-sys/build.rs
tgonzalezorlandoarm: Sorry, what I meant to say is "in the rest of the function call", my bad.

tgonzalezorlandoarm (approved):
LGTM, Thanks!

psa-crypto-sys/build.rs
```diff
@@ -66,6 +66,73 @@ mod common {
...
+                .map(|x: PathBuf| -> String { x.into_os_string().into_string().unwrap() })
+                .collect();
+            let include_dir = include_dirs.join(" ");
+
+            return Ok(include_dir);
+        }
+
+        Err(Error::new(
+            ErrorKind::Other,
+            "interface feature necessitates MBEDTLS_INCLUDE_DIR environment variable",
```
gowthamsk-arm, line 132: Should we also include `or library discoverable by pkg-config` (or something along similar lines) in the error log here?

gowthamsk-arm:
@billatarm how to test the pkg-config feature locally? The CI doesn't do it right now so want to test it manually. 
Installing mbedtls with your pkg-config patch and then building this crate, is it sufficient? :)

psa-crypto-sys/build.rs
billatarm: No, this is for the interface feature which doesn't need the libdir, just the headers.

gowthamsk-arm:
Thanks for the steps @billatarm 
I just tested the pkg-config changes with the mbedtls PR that you put up and it works all well.

gowthamsk-arm (approved):
LGTM

tgonzalezorlandoarm (approved):

PR #141, opened 2024-01-31
nightly/audit: Temporary ignore RUSTSEC-2024-0006

tgonzalezorlandoarm:
shlex is currently only being brought by bindgen, which does not use the concerning APIs except for testing.

This needs spiffe to be upgraded to be finally be resolved, which will not happen now.


gowthamsk-arm:
Can you add a small comment on top of the ignore command?

gowthamsk-arm (approved):
LGTM

PR #143, opened 2024-03-04
Enabled psa_mac_compute and psa_mac_verify functions.

mes5k:
This PR enables the `psa_mac_compute` and `psa_mac_verify` functions.  I added new tests, fixed doc tests and formatted the code.

Everything seems to work fine, at least to my inexperienced eye. Please review closely and let me know if anything additional is needed.

My only real conceptual question is whether I should be checking if the size > 0 after the called to `psa_crypto_sys::PSA_MAC_LENGTH` in the `mac_length` function.  This makes sense to me as a belt and suspenders approach to ensuring we don't end up with a 0 size mac buffer.  On the other hand, I haven't been able to come up with a test scenario where this might happen, so maybe it's not necessary? 

Thanks!

### DIFF
--- a/psa-crypto/src/operations/mac.rs
+++ b/psa-crypto/src/operations/mac.rs
@@ -4,34 +4,32 @@
 //! # Message Authentication Code (MAC) operations
 
 use crate::initialized;
-use crate::types::key::Id;
 use crate::types::algorithm::Mac;
-use crate::types::status::{Result, Status, Error};
-
+use crate::types::key::Id;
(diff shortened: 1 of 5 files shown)


mes5k:
Thanks @hug-dev, just pushed the updates.

hug-dev (approved):
Thanks for this! Nice formatting and tests too 👍

hug-dev (approved):
Thank you!

PR #145, opened 2025-11-16
build: make psa-crypto-sys no_std, fix zeroize with no_std

dsseng:
As psa-crypto-sys does not rely on any std features, it should be safe
to completely switch its bindgen options to use core, and mark it as
no_std. This allows this crate to be used in embedded applications such
as TrustedFirmware-M Secure Partitions.

zeroize should not need `alloc` default feature, as this crate does not
make use of Vec or Box and does not derive zeroize on them. Disable all
default features to avoid unnecessary dependency on std.

Tested to provide access to hash functions via `operations` feature in
a TF-M Secure Partition sample.

Signed-off-by: Dmitrii Sharshakov <d3dx12.xx@gmail.com>

### DIFF
--- a/psa-crypto-sys/build.rs
+++ b/psa-crypto-sys/build.rs
@@ -171,6 +171,7 @@ mod common {
             .clang_arg(format!("-I{mbed_include_dir}"))
             .header("src/c/shim.h")
             .blocklist_type("max_align_t")
+            .use_core()
             .generate_comments(false)
             .size_t_is_usize(true);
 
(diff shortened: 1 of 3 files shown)


hug-dev:
Please feel free to continue working on this if interested, I will review and merge it myself :)

dsseng:
Will rebase this evening, thank you

hug-dev (approved):
Awesome, thank you!

PR #148, opened 2026-04-27
upgrade base64 to 0.22.1

alexanderkjall:
based on: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/psa-crypto/debian/patches/relax-dep.diff?ref_type=heads


hug-dev (approved):
Thank you!

PR #149, opened 2026-04-30
Bump bindgen crate version

anta5010:


psa-crypto-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "psa-crypto-sys"
-version = "0.12.0"
+version = "0.13.0"
 authors = ["Parsec Project Contributors"]
 edition = "2018"
```
ionut-arm, line 5: Should we change edition too?

ionut-arm (approved):
Looks good!

psa-crypto-sys/Cargo.toml
anta5010: I don't know. Should we? Our code and the dependency crate don't require new edition.
What's the recommendation from Rust people?

